PR #1, opened 2020-02-10
Add Role support

brinxmat:


src/main/java/no/unit/nva/model/Role.java
```diff
@@ -0,0 +1,34 @@
...
+     *
+     * @param value value
+     * @return  enum
+     */
+    public static Role lookup(String value) {
+        return Arrays
+                .stream(values())
+                .filter(nameType -> nameType.getValue().equalsIgnoreCase(value))
+                .findAny()
+                .orElseThrow(IllegalArgumentException::new);
```
janbj: What about
`.orElseThrow(() -> new IllegalArgumentException(format("%s not a valid Role, expected one of: %s", value, stream(Role.values()).map(Role::toString).collect(joining(", ")))));
`

PR #2, opened 2020-03-03
Np 399 update datamodel

janbj:


src/main/java/no/unit/nva/model/Contributor.java
```diff
@@ -1,17 +1,14 @@
 package no.unit.nva.model;
 
-import java.net.URI;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
```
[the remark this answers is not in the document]
janbj: Orestis said no

src/main/java/no/unit/nva/model/Contributor.java
```diff
@@ -20,53 +17,26 @@ public Contributor() {
...
 
-    public void setNameType(String nameType) {
-        this.nameType = nameType;
+    public void setIdentity(Identity identity) {
+        this.identity = identity;
     }
 
-    public List<String> getImportedAffiliation() {
-        return importedAffiliation;
+    public List<Organization> getAffiliation() {
```
brinxmat: If affiliations are always plural, we should make this explicit in the nomenclature

src/main/java/no/unit/nva/model/EntityDescription.java
```diff
@@ -1,14 +1,18 @@
...
+
+import java.net.URI;
 import java.util.List;
 import java.util.Map;
 
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
 public class EntityDescription {
 
-    private String type;
+    private String publicationType;
```
brinxmat: I suspect publication type is not a string, there are a limited set of possibilities, we should speak with PO, for now JournalArticle is the only type.

src/main/resources/publicationContext.json
```diff
@@ -0,0 +1,27 @@
+{
+  "@vocab": "https://example.org/cheeky#",
```
brinxmat: Probably not.

src/main/resources/publicationFrame.json
```diff
@@ -0,0 +1,38 @@
+{
+  "@context": {
+    "@vocab": "https://example.org/cheeky#",
```
brinxmat: Again :)

src/test/java/no/unit/nva/model/util/ContextUtilTest.java
```diff
@@ -0,0 +1,24 @@
...
+public class ContextUtilTest {
+
+    @Test
+    public void testInjectContextIntoJsonNodeNotAnObject() {
+
+        ObjectMapper objectMapper = new ObjectMapper();
+        ArrayNode arrayNode = objectMapper.createArrayNode();
+        ObjectNode objectNode = objectMapper.createObjectNode();
+
+        Assertions.assertThrows(IllegalArgumentException.class, () -> {
```
brinxmat: Do you want to test the message that is returned?

src/main/resources/publicationContext.json
janbj: What should we use then?

src/main/java/no/unit/nva/model/EntityDescription.java
janbj: Yes, I tried using the enum but I couldn't guarantee all possible mappings when Terje was testing various DOIs

src/main/resources/publicationFrame.json
janbj: Sea otter

src/main/java/no/unit/nva/model/EntityDescription.java
brinxmat: This got me to thinking about datacite and crossref, we should talk with JEG

PR #3, opened 2020-03-17
OrgNumberMapper and fields for indexing

janbj:


src/test/java/no/unit/nva/model/util/OrgNumberMapperTest.java
```diff
@@ -0,0 +1,29 @@
...
+import static no.unit.nva.model.util.OrgNumberMapper.NO_MAPPING_ERROR_MESSAGE;
+
+public class OrgNumberMapperTest {
+
+    public static final String VALID_CRISTIN_ID = "https://api.cristin.no/v2/institutions/20202";
+    public static final String VALID_ORG_NUMBER = "919477822";
+    public static final String INVALID_ORG_NUMBER = "123";
+
+    @Test
+    public void testOrgNumberExistingInMapReturnsCristinId() {
```
axthosarouris: the "test" prefix is not necessary. We are in a test class with the `@Test` tag over the method

src/main/java/no/unit/nva/model/util/OrgNumberMapper.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.Map;
+
+public class OrgNumberMapper {
+
+    public static final Map<String,URI> orgNumberCristinIdMap;
+    public static final String NO_MAPPING_ERROR_MESSAGE = "No mapping for orgNumber: ";
+
+    static {
+        orgNumberCristinIdMap = new HashMap<>();
+        orgNumberCristinIdMap.put("919477822", URI.create("https://api.cristin.no/v2/institutions/20202"));
```
axthosarouris: Can you replace these strings with constants with a descriptive name?

src/main/java/no/unit/nva/model/util/OrgNumberMapper.java
```diff
@@ -0,0 +1,32 @@
...
+        orgNumberCristinIdMap.put("914086434", URI.create("https://api.cristin.no/v2/units/7467.0.0.0"));
+        orgNumberCristinIdMap.put("952125001", URI.create("https://api.cristin.no/v2/units/258.0.0.0"));
+    }
+
+    /**
+     * Map OrgNumber to CristinId.
+     *
+     * @param orgNumber orgNumber
```
axthosarouris: I understand that we have a rule to create Javadocs which is good, but we seem to misuse it. I am not referring necessarily to this example but the pattern `@param orgNumber orgNumber` is frequent. Maybe we should reconsider this rule? No strong opinions about it at the moment. 
(This is a just a thought. No need for changing code here.)

src/main/java/no/unit/nva/model/Publication.java
```diff
@@ -44,6 +45,33 @@ private Publication(Builder builder) {
...
+            () -> new IllegalStateException("Property publisher.id can not be null")).toString();
+    }
+
+    /**
+     * Getter to create value used in DynamoDB indexing during serialization.
+     *
+     * @return publisherOwnerDate
+     */
+    public String getPublisherOwnerDate() {
+        return String.join("#",
```
axthosarouris: Could this be the "DynamoKeyDelimiter"?

src/main/java/no/unit/nva/model/util/OrgNumberMapper.java
```diff
@@ -0,0 +1,40 @@
+package no.unit.nva.model.util;
+
+import java.net.URI;
+import java.util.HashMap;
+import java.util.Map;
+
+public class OrgNumberMapper {
+
```
brinxmat: Could we use an enum for this?

PR #4, opened 2020-03-17
Np 631 add abstract to entity data

brinxmat:


src/test/java/no/unit/nva/model/PublicationTest.java
```diff
@@ -115,6 +112,7 @@ private EntityDescription getEntityDescription() {
                 .withDate(getPublicationDate())
                 .withPublicationType(PublicationType.JOURNAL_ARTICLE)
                 .withContributors(Collections.singletonList(getContributor()))
+                .withAbstract("En lang streng som beskriver innholdet i dokumentet metdataene omtaler.")
```
axthosarouris: Not important, but shouldn't that be English?

brinxmat: The publication language is set to Norwegian, so the abstract must match.

PR #5, opened 2020-03-17
Add tags to entity description

brinxmat:


src/main/java/no/unit/nva/model/EntityDescription.java
```diff
@@ -114,8 +116,17 @@ public void setAbstract(String mainLanguageAbstract) {
...
+        return tags;
+    }
+
+    public void setTags(List<String> tags) {
+        this.tags = tags;
+    }
+
     public static final class Builder {
         public String mainLanguageAbstract;
+        public List<String> tags;
```
janbj: Just noticed that the other fields in the builder are private

PR #6, opened 2020-03-17
Add description property

brinxmat:


src/main/java/no/unit/nva/model/EntityDescription.java
```diff
@@ -114,8 +116,17 @@ public void setAbstract(String mainLanguageAbstract) {
...
+        return description;
+    }
+
+    public void setDescription(String description) {
+        this.description = description;
+    }
+
     public static final class Builder {
         public String mainLanguageAbstract;
+        public String description;
```
janbj: Use private instead

src/test/java/no/unit/nva/model/PublicationTest.java
```diff
@@ -113,6 +113,7 @@ private EntityDescription getEntityDescription() {
                 .withPublicationType(PublicationType.JOURNAL_ARTICLE)
                 .withContributors(Collections.singletonList(getContributor()))
                 .withAbstract("En lang streng som beskriver innholdet i dokumentet metdataene omtaler.")
```
janbj: At some point we need to fix the type error in "metadataene" :see_no_evil:

PR #7, opened 2020-03-18
Add NPI subject heading

brinxmat:


src/main/java/no/unit/nva/model/EntityDescription.java
```diff
@@ -114,8 +116,17 @@ public void setAbstract(String mainLanguageAbstract) {
...
+        return npiSubjectHeading;
+    }
+
+    public void setNpiSubjectHeading(String npiSubjectHeading) {
+        this.npiSubjectHeading = npiSubjectHeading;
+    }
+
     public static final class Builder {
         public String mainLanguageAbstract;
+        public String npiSubjectHeading;
```
janbj: Use private instead

PR #8, opened 2020-03-18
Np 641 add journal oriented single fields to entity description

brinxmat:


src/main/java/no/unit/nva/model/EntityDescription.java
```diff
@@ -148,7 +151,16 @@ public void setDescription(String description) {
...
+    public JournalReference getJournalReference() {
+        return journalReference;
+    }
+
+    public void setJournalReference(JournalReference reference) {
+        this.journalReference = reference;
+    }
+
     public static final class Builder {
+        public JournalReference journalReference;
```
janbj: private

src/main/java/no/unit/nva/model/File.java
```diff
@@ -24,6 +28,9 @@ private File(Builder builder) {
         setMimeType(builder.mimeType);
         setSize(builder.size);
         setLicense(builder.license);
+        setAdministrativeAgreement(builder.administrativeAgreement);
+        setPublisher(builder.publisherAuthority);
```
janbj: publisher vs publisherAuthority?

src/main/java/no/unit/nva/model/JournalReference.java
```diff
@@ -0,0 +1,90 @@
...
+    public PublicationInstance getPublicationInstance() {
+        return publicationInstance;
+    }
+
+    public void setPublicationInstance(PublicationInstance publicationInstance) {
+        this.publicationInstance = publicationInstance;
+    }
+
+    public static final class Builder {
+        public PublicationInstance publicationInstance;
```
janbj: private

src/main/java/no/unit/nva/model/PublicationContext.java
```diff
@@ -0,0 +1,110 @@
...
+    public boolean isPeerReviewed() {
+        return peerReviewed;
+    }
+
+    public void setPeerReviewed(boolean peerReviewed) {
+        this.peerReviewed = peerReviewed;
+    }
+
+    public static class Builder {
+        public String name;
```
janbj: Again, should probably be private

src/main/java/no/unit/nva/model/File.java
```diff
@@ -66,28 +73,34 @@ public void setLicense(License license) {
         this.license = license;
     }
 
-    @Override
-    public boolean equals(Object o) {
```
axthosarouris: I think deep equals in the data-model is necessary. We should bear in mind that we will potentially need to compare equivalent instances and we will need deep equals.

src/main/java/no/unit/nva/model/File.java
```diff
@@ -66,28 +73,34 @@ public void setLicense(License license) {
...
+    public Instant getEmbargoDate() {
+        return embargoDate;
+    }
+
+    public void setEmbargoDate(Instant embargoDate) {
+        this.embargoDate = embargoDate;
     }
 
     public static final class Builder {
+        public boolean administrativeAgreement;
```
axthosarouris: Maybe put the Builder class at the end of the parent class for clarity

PR #9, opened 2020-03-24
Add support for projects, and associated grants, approvals

brinxmat:
As it turns out, project is more complex that at first imagined, while we actually need a reference, we also need to be able to serialise data about projects, including:

- Grants
- Approvals

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,7 +61,7 @@ local.properties
 
 ### Eclipse Patch ###
 # Eclipse Core
-.researchProject
+.project
 
 # JDT-specific (Eclipse Java Development Tools)
 .classpath
@@ -114,7 +114,7 @@ cmake-build-*/
 # Mongo Explorer plugin
 .idea/**/mongoSettings.xml
 
-# File-based researchProject format
+# File-based project format
 *.iws
 
 # IntelliJ
@@ -272,7 +272,7 @@ gradle-app.setting
 # Avoid ignoring Gradle wrapper jar file (.jar files are usually ignored)
 !gradle-wrapper.jar
 
-# Cache of researchProject
+# Cache of project
 .gradletasknamecache
 
 # # Work around https://youtrack.jetbrains.com/issue/IDEA-116898
--- a/.gitignore
+++ b/.gitignore
@@ -40,7 +40,7 @@ local.properties
 .target
 
 # Tern plugin
-.tern-project
+.tern-researchProject
 
 # TeXlipse plugin
 .texlipse
@@ -61,7 +61,7 @@ local.properties
 
 ### Eclipse Patch ###
 # Eclipse Core
-.project
+.researchProject
 
 # JDT-specific (Eclipse Java Development Tools)
 .classpath
@@ -114,7 +114,7 @@ cmake-build-*/
 # Mongo Explorer plugin
 .idea/**/mongoSettings.xml
 
-# File-based project format
+# File-based researchProject format
 *.iws
 
 # IntelliJ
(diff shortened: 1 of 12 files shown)


.gitignore
```diff
@@ -40,7 +40,7 @@ local.properties
 .target
 
 # Tern plugin
-.tern-project
+.tern-researchProject
```
janbj: Did you mean to rename all of these?

src/main/java/no/unit/nva/model/Approval.java
```diff
@@ -0,0 +1,112 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+import java.time.Instant;
+import java.util.Objects;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class Approval {
+    private Instant date;
+    private String approvedBy;
+    private String approvalStatus;
```
janbj: No enum?

src/main/java/no/unit/nva/model/Publication.java
```diff
@@ -9,10 +9,11 @@
 import java.util.UUID;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+@SuppressWarnings("PMD.ExcessivePublicCount")
 public class Publication {
 
-    public static final String DYNAMODB_KEY_DELIMITER = "#";
-    
+    private static final String DYNAMODB_KEY_DELIMITER = "#";
```
janbj: I see no problem with constants being public

src/main/java/no/unit/nva/model/Publication.java
```diff
@@ -178,117 +181,143 @@ public void setFileSet(FileSet fileSet) {
...
+        }
+
+        public Builder withStatus(PublicationStatus val) {
+            status = val;
             return this;
         }
 
-        public Builder withStatus(PublicationStatus status) {
-            this.status = status;
+        public Builder withOwner(String val) {
```
janbj: Not a fan of argument name here, is there a setting in the plugin you forgot to enable?

.gitignore
brinxmat: Probably not :D

.gitignore
```diff
@@ -61,7 +61,7 @@ local.properties
 
 ### Eclipse Patch ###
 # Eclipse Core
-.project
+.researchProject
```
janbj: Just project

.gitignore
```diff
@@ -114,7 +114,7 @@ cmake-build-*/
 # Mongo Explorer plugin
 .idea/**/mongoSettings.xml
 
-# File-based project format
+# File-based researchProject format
```
janbj: Just project

.gitignore
```diff
@@ -272,7 +272,7 @@ gradle-app.setting
 # Avoid ignoring Gradle wrapper jar file (.jar files are usually ignored)
 !gradle-wrapper.jar
 
-# Cache of project
+# Cache of researchProject
```
janbj: Just project

src/main/java/no/unit/nva/model/Publication.java
```diff
@@ -178,57 +181,79 @@ public void setFileSet(FileSet fileSet) {
...
+        this.project = project;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) {
             return true;
         }
-        if (o == null || getClass() != o.getClass()) {
+        if (!(o instanceof Publication)) {
```
axthosarouris: why is the null check removed?

brinxmat: This is generated code, from what I can see, the null-check was only in place because the later call to getClass would throw an NPE. The generated code now obviates the need for the check because of the use of instanceof.

PR #13, opened 2020-03-28
Np 696 add contributor email

brinxmat:
This commit builds on NP-695-add-contributor-corresponding-author, and should be merged into that branch.

We add:

- Contributor.email
- MalformedContributorException
- constructor Contributor(Builder builder) is modified to check that the email has been set, or returns a MalformedContributorException with appropriate message
- add functionality to builder
- add tests to ensure that Contributors are well formed
- update Publication tests with new field
- update JSON-LD framing

### DIFF
--- a/src/main/java/no/unit/nva/model/Contributor.java
+++ b/src/main/java/no/unit/nva/model/Contributor.java
@@ -24,8 +24,8 @@ public class Contributor {
 
     }
 
-    private Contributor(Builder builder) {
-        if (builder.correspondingAuthor && (isNull(builder.email) || builder.email.isBlank())) {
+    private Contributor(Builder builder) throws MalformedContributorException {
+        if (isCorrespondAuthorWithEmail(builder)) {
             throw new MalformedContributorException(CORRESPONDING_AUTHOR_EMAIL_MISSING);
         }
         setIdentity(builder.identity);
@@ -34,6 +34,10 @@ public class Contributor {
         setSequence(builder.sequence);
         setCorrespondingAuthor(builder.correspondingAuthor);
         setEmail(builder.email);
+    }
+
+    private boolean isCorrespondAuthorWithEmail(Builder builder) {
+        return builder.correspondingAuthor && (isNull(builder.email) || builder.email.isBlank());
     }
 
     public Identity getIdentity() {
@@ -152,7 +156,7 @@ public class Contributor {
             return this;
         }
 
-        public Contributor build() {
+        public Contributor build() throws MalformedContributorException {
             return new Contributor(this);
         }
     }
--- a/src/main/java/no/unit/nva/model/Contributor.java
+++ b/src/main/java/no/unit/nva/model/Contributor.java
@@ -1,5 +1,7 @@
 package no.unit.nva.model;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import no.unit.nva.model.exceptions.MalformedContributorException;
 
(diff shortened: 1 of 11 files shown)


src/main/java/no/unit/nva/model/Contributor.java
```diff
@@ -1,27 +1,39 @@
...
     private Integer sequence;
+    private boolean correspondingAuthor;
+    private String email;
 
     public Contributor() {
 
     }
 
     private Contributor(Builder builder) {
+        if (builder.correspondingAuthor && (isNull(builder.email) || builder.email.isBlank())) {
```
axthosarouris: refactor the condition in a method with name correspondingAuthorMailIsIsValid

src/main/java/no/unit/nva/model/exceptions/MalformedContributorException.java
```diff
@@ -0,0 +1,7 @@
+package no.unit.nva.model.exceptions;
+
+public class MalformedContributorException extends RuntimeException {
```
axthosarouris: why unchecked? Unchecked exceptions are a bit less visible

brinxmat: Updated

src/main/java/no/unit/nva/model/Contributor.java
```diff
@@ -1,27 +1,43 @@
...
 import java.util.List;
 import java.util.Objects;
 
+import static java.util.Objects.isNull;
+
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 public class Contributor {
 
+    public static final String CORRESPONDING_AUTHOR_EMAIL_MISSING =
+            "The Contributor is corresponding author, but no email for correspondence is set";
```
janbj: So email may be set for all contributors but is obligatory for corresponding authors?

src/main/java/no/unit/nva/model/Contributor.java
```diff
@@ -1,27 +1,43 @@
...
+    private boolean correspondingAuthor;
+    private String email;
 
     public Contributor() {
 
     }
 
-    private Contributor(Builder builder) {
+    private Contributor(Builder builder) throws MalformedContributorException {
+        if (isCorrespondAuthorWithEmail(builder)) {
```
janbj: I have a suggestion, make a Constructor annotated with JsonCreator that has all obligatory fields and move the validation there. We want the validation to be active for normal deserialization too, not just for the Builder's Constructor.

src/main/java/no/unit/nva/model/exceptions/MalformedContributorException.java
```diff
@@ -0,0 +1,7 @@
+package no.unit.nva.model.exceptions;
+
+public class MalformedContributorException extends Exception {
```
janbj: Do you want this very specific exception to be handled locally during stream processing? If yes then Exception is ok, but if you want it to bubble up to main exception handler then inherit from RuntimeException.

src/main/java/no/unit/nva/model/exceptions/InvalidNpiLevelException.java
```diff
@@ -1,6 +1,6 @@
 package no.unit.nva.model.exceptions;
 
-public class InvalidNpiLevelException extends RuntimeException {
+public class InvalidNpiLevelException extends Exception {
```
janbj: See comment for MalformedContributorException

src/test/java/no/unit/nva/model/ContributorTest.java
```diff
@@ -0,0 +1,92 @@
...
+                .build();
+    }
+
+    private static Identity getIdentity() {
+        return new Identity.Builder().withName("Smith, Kim").build();
+    }
+
+    @DisplayName("Contributor throws MalformedContributorException when corresponding author, but no email is set")
+    @Test
+    void contributorThrowsExceptionWhenCorrespondingAuthorNoEmail() {
```
janbj: This only works when using the Builder.

src/main/java/no/unit/nva/model/Contributor.java
brinxmat: Yes

src/main/java/no/unit/nva/model/exceptions/MalformedContributorException.java
brinxmat: We need to have a very concrete relationship with the exceptions here, I guess throwing them and throwing them again leaves a clearer trail than an unchecked exception. We need to discuss this further.

janbj: Exceptions that we know how to handle should be checked, the rest should generally be unchecked.

axthosarouris: Sending a HTTP code is a way of handling, so if we anticipate that something can wrong, the it should be a checked Exception. Not handling is equivalent with returning Internal Server Error

PR #14, opened 2020-03-28
Add correspondingAuthor

brinxmat:
Here, we update the contributor object with a new field, correspondingAuthor, which contains information about whether a given contributor is to be understood as the corresponding author (that is to say the author that communicated with reviewers, publishers, etc.) This is a boolean value.

- new field Contributor.correspondingAuthor, boolean
- update builder, equals, hash
- update tests
- update JSON-LD frame
- removed java 8 from travis jobs

### DIFF
--- a/src/main/java/no/unit/nva/model/Contributor.java
+++ b/src/main/java/no/unit/nva/model/Contributor.java
@@ -58,6 +58,14 @@ public class Contributor {
         this.role = role;
     }
 
+    public boolean isCorrespondingAuthor() {
+        return correspondingAuthor;
+    }
+
+    public void setCorrespondingAuthor(boolean correspondingAuthor) {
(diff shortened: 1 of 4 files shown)


src/main/java/no/unit/nva/model/Contributor.java
```diff
@@ -61,26 +63,40 @@ public boolean equals(Object o) {
...
     public int hashCode() {
-        return Objects.hash(getIdentity(), getAffiliations(), getRole(), getSequence());
+        return Objects.hash(getIdentity(),
+                getAffiliations(),
+                getRole(),
+                getSequence(),
+                isCorrespondingAuthor());
+    }
+
+    public boolean isCorrespondingAuthor() {
```
axthosarouris: Group all setters and getters together

PR #23, opened 2020-04-30
NP-831: Publication reference contains URI

axthosarouris:


src/main/java/no/unit/nva/model/Reference.java
```diff
@@ -2,13 +2,15 @@
...
+import nva.commons.utils.JacocoGenerated;
 
+import java.net.URI;
 import java.util.Objects;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 public class Reference {
     private PublicationContext publicationContext;
-    private String doi;
+    private URI doi;
```
brinxmat, line 13: We are now certain that a doi is a URI in all cases?

axthosarouris: Any reference in doi in datamodel is now a URI

PR #24, opened 2020-05-06
Add interfaces, fix handling of Page

brinxmat:


src/test/java/no/unit/nva/model/instancetypes/PublicationInstanceTest.java
```diff
@@ -37,10 +40,10 @@ void publicationInstanceReturnsSerializedJsonWhenValidRangeIsInput() throws Json
                 objectMapper.writerWithDefaultPrettyPrinter().writeValueAsString(publicationInstance));
     }
 
-    @DisplayName("Publication instance can be serialized with Range object")
+    @DisplayName("Publication instance can ONLY be serialized with Range object")
```
axthosarouris, line 43: "Publication instance cannot be serialized when pages is not a Range object."

src/main/java/no/unit/nva/model/instancetypes/JournalArticle.java
```diff
@@ -99,12 +128,12 @@ public boolean equals(Object o) {
         if (!(o instanceof JournalArticle)) {
             return false;
         }
-        JournalArticle journalArticle = (JournalArticle) o;
-        return Objects.equals(getVolume(), journalArticle.getVolume())
-                && Objects.equals(getIssue(), journalArticle.getIssue())
-                && Objects.equals(getArticleNumber(), journalArticle.getArticleNumber())
-                && Objects.equals(getPages(), journalArticle.getPages())
-                && isPeerReviewed() == journalArticle.isPeerReviewed();
+        JournalArticle that = (JournalArticle) o;
```
janbj, line 131: What settings do you use when generating this?

src/test/resources/publication_instance_monograph_pages.json
```diff
@@ -1,5 +1,5 @@
 {
-  "type" : "PublicationInstance",
+  "type" : "JournalArticle",
```
janbj, line 2: Filename is still publication_instance, is that intentional?

PR #26, opened 2020-05-14
Np 905 publication interfaces

janbj:


src/test/java/no/unit/nva/PublicationMapperTest.java
```diff
@@ -0,0 +1,81 @@
...
+import org.junit.jupiter.api.Test;
+
+public class PublicationMapperTest {
+
+    public static final URI SOME_URI = URI.create("http://example.org");
+    public static final String SOME_OWNER = "owner";
+    public static final ObjectNode SOME_CONTEXT = objectMapper.createObjectNode();
+
+    @Test
+    public void canMapCreatePublicationRequestToNewPublication() throws Exception {
```
axthosarouris, line 29: can map create publication request to new publication? 
can map create PublicationRequest to new Publication?
can MapCreate PublicationRequest to NewPublication?

try to keep the format methodReturnsOutputWhenInputIsSomething

axthosarouris, line 29: we should agree on a test name pattern.

build.gradle
```diff
@@ -26,7 +26,7 @@ def jacksonVersion = '2.10.2'
 gradleLint.rules = ['unused-dependency']
 
 dependencies {
-    implementation group: 'com.github.BIBSYSDEV', name: 'nva-commons', version: 'v0.1.6'
+    implementation group: 'com.github.BIBSYSDEV', name: 'nva-commons', version: '0.3.3'
```
axthosarouris, line 29: this should have been its own branch, create a PR for this, and then branch out from that branch to continue your work.

src/test/java/no/unit/nva/PublicationMapperTest.java
janbj: Sorry, my fault - I find the one we already agreed upon to be very cumbersome.

axthosarouris:
The code seems fine. I understand what you want to do but I missing the scope and the context. I think we are very bad in making correct branches and PRs. We do not have a narrow scope and we do not describe what is the purpose of the PR in the PR description. We (including me) dump a bunch of different changes to the others and we  pay no attention of keeping a narrow scope and describing the context of the scope in the description of the PR. On top of that we do not name the tests correctly and that makes things much more difficult. 

Big PRs with no scope have many side-effects:
1. People get tired of reading long PRs and just skim through.
2. Reviewers do not understand what is the exact purpose of the PR
3. The person who made the changes does not keep track of the scope of the changes.
4. The comments that we are exchanging are only comments about the programming style. Although useful we will end up being excellent programmers creating bad services.

PR #29, opened 2020-05-19
NP-884 Added publication fields doi and doiRequest

nraanes:


src/main/resources/publicationFrame.json
```diff
@@ -46,6 +46,7 @@
     "isOpenAccess": "openAccess",
     "isPeerReviewed": "peerReviewed",
     "hasDoi": "doi",
+    "hasDoiRequest": "doiRequest",
```
nraanes, line 49: I was not sure about adding this line, but it seemed appropriate.

src/main/java/no/unit/nva/model/DoiRequest.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Objects;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class DoiRequest {
+
+    private DoiRequestStatus status;
+    private Instant date;
+
+    public DoiRequest() {
+
```
axthosarouris, line 16: space

src/main/java/no/unit/nva/model/DoiRequestStatus.java
```diff
@@ -0,0 +1,7 @@
+package no.unit.nva.model;
+
+public enum DoiRequestStatus {
+    REQUESTED,
+    ACCEPTED,
```
brinxmat, line 5: APPROVED?

src/main/resources/publicationFrame.json
brinxmat: Yes, it is for the mapping to the external model

PR #33, opened 2020-05-20
New property for indexing doiRequest

janbj:
Added getDoiRequestStatusDate to be used for indexing


src/main/java/no/unit/nva/WithIndex.java
```diff
@@ -34,4 +36,21 @@ default String getPublisherOwnerDate() {
         );
     }
 
+    /**
+     * Getter to create value used in DynamoDB indexing during serialization.
+     *
+     * @return doiRequestStatusDate
+     */
+    default String getDoiRequestStatusDate() {
```
axthosarouris, line 44: I would suggest to create a variable for each argument and to put the toString in a map() function before the .orElseThrow. In that way, you have the elseThrow as a last command and not in the middle.

Declaring a one-time variable does not affect efficiency, Compilers recognize these patterns and remove them during compilation time.

Suggestion:

```
public interface WithIndex extends WithInternal {

    String DYNAMODB_KEY_DELIMITER = "#";

    /**
     * Getter to create value used in DynamoDB indexing during serialization.
     *
     * @return publisherId
     */
    default String getPublisherId() {
        Organization publisher = Optional.ofNullable(getPublisher())
                .orElseThrow(() -> new IllegalStateException("Object publisher can not be null"));
        return Optional.ofNullable(publisher.getId())
                .map(URI::toString)
                .orElseThrow(() -> new IllegalStateException("Property publisher.id can not be null"));
    }

    /**
     * Getter to create value used in DynamoDB indexing during serialization.
     *
     * @return publisherOwnerDate
     */
    default String getPublisherOwnerDate() {
        String owner = Optional.ofNullable(getOwner())
                .orElseThrow(() -> new IllegalStateException("Property owner can not be null"));
        String date =Optional.ofNullable(getModifiedDate())
                .map(Instant::toString)
                .orElseThrow(() -> new IllegalStateException("Property modifiedDate can not be null"));
        return String.join(DYNAMODB_KEY_DELIMITER, getPublisherId(),owner,date);
    }

    /**
     * Getter to create value used in DynamoDB indexing during serialization.
     *
     * @return doiRequestStatusDate
     */
    default String getDoiRequestStatusDate() {
        if (isNull(getDoiRequest())) {
            return null; // not indexed
        }
        String statusString =Optional.ofNullable(getDoiRequest().getStatus()).map(Enum::toString)
                .orElseThrow(() -> new IllegalStateException("Property doiRequest.status can not be null"));
        String dateString = Optional.ofNullable(getDoiRequest().getDate()).map(Instant::toString)
                .orElseThrow(() -> new IllegalStateException("Property doiRequest.date can not be null"));

        return String.join(DYNAMODB_KEY_DELIMITER,statusString,dateString);

    }

}

```

PR #35, opened 2020-05-22
Add type book, tests and update PublicationContext JSON subtype handling

brinxmat:
- Add Class Book
- Add BookTest
- update JSON subtype handling in PublicationContext

### DIFF
--- a/src/main/java/no/unit/nva/model/contexttypes/Book.java
+++ b/src/main/java/no/unit/nva/model/contexttypes/Book.java
@@ -1,0 +1,114 @@
+package no.unit.nva.model.contexttypes;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import no.unit.nva.model.Level;
+import nva.commons.utils.JacocoGenerated;
+
+import java.util.List;
+import java.util.Objects;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class Book implements PublicationContext {
+
+    private String title;
+    private Level level;
+    private boolean openAccess;
+    private boolean peerReviewed;
+    @JsonProperty("isbn")
+    private List<String> isbns;
+
+    /**
+     * Constructor for Book objects.
+     * @param title The title of the book.
+     * @param level The NPI level of the publisher.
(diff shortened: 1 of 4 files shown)


src/main/java/no/unit/nva/model/contexttypes/Book.java
```diff
@@ -0,0 +1,114 @@
...
+                @JsonProperty("peerReviewed") boolean peerReviewed,
+                @JsonProperty("isbn") List<String> isbns) {
+        this.title = title;
+        this.level = level;
+        this.openAccess = openAccess;
+        this.peerReviewed = peerReviewed;
+        this.isbns = isbns;
+    }
+
+    @JsonCreator
```
axthosarouris, line 42: misplaced tag. it should be above the constructor

src/test/java/no/unit/nva/model/contexttypes/BookTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import java.nio.file.Path;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class BookTest {
```
axthosarouris, line 18: I am wondering whether it would be a good idea to have two methods `fromJson(String json)` (static) and `toJson()` that would also include  serialization error handling. But this maybe should be a different PR.

src/test/java/no/unit/nva/model/contexttypes/BookTest.java
```diff
@@ -0,0 +1,64 @@
...
+    @DisplayName("Book can deserialize a book")
+    @Test
+    void objectMapperReturnsBookWhenInputIsValidJson() throws JsonProcessingException {
+        String expectedType = "Book";
+        String expectedTitle = "The history of JSON in this project";
+        Level expectedLevel = Level.LEVEL_2;
+        String expectedIsbn = "123 123123 1112";
+        String json = generateJsonFromTemplate(expectedType, expectedTitle, expectedLevel,
+                true, true, expectedIsbn);
+        Book book = objectMapper.readValue(json, Book.class);
```
axthosarouris, line 33: Suggestion: Create an expected book instance and do assertEquals(expectedBook, actualBook)?
If you also refactor the generation of an expected book to a method, you get an (almost) free test on equals as well :)

PR #36, opened 2020-05-24
Add series support to Book

brinxmat:
In this PR, because Report is simply a subtype of Book, we need Series:
- Add series concept, including
  - Series, including Title and Number
  - NullSeries object for cases where a Book is not part a series
- Refactor tests

### DIFF
--- a/src/main/java/no/unit/nva/model/NullSeries.java
+++ b/src/main/java/no/unit/nva/model/NullSeries.java
@@ -1,0 +1,9 @@
+package no.unit.nva.model;
+
+public class NullSeries implements Series {
+
+    @Override
+    public String getTitle() {
+        return null;
+    }
+}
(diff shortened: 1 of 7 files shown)


src/main/java/no/unit/nva/model/contexttypes/Book.java
```diff
@@ -0,0 +1,155 @@
...
+    private String title;
+    private Series series;
+    private Level level;
+    private boolean openAccess;
+    private boolean peerReviewed;
+    @JsonProperty("isbn")
+    private List<String> isbns;
+
+    public Book() {
+
```
axthosarouris, line 27: redundant line

src/main/java/no/unit/nva/model/NullSeries.java
```diff
@@ -0,0 +1,9 @@
+package no.unit.nva.model;
+
+public class NullSeries implements Series {
```
janbj, line 3: Deserialized it will look like this, but I guess it is okay?
`...
"series": { "type": "NullSeries" }
...
`

PR #37, opened 2020-05-24
Np 923 create type report add report

brinxmat:
- Create interface SerialPublication
- Create Utility for IssnChecking
- Add Report

### DIFF
--- a/src/main/java/no/unit/nva/model/contexttypes/Book.java
+++ b/src/main/java/no/unit/nva/model/contexttypes/Book.java
@@ -13,7 +13,7 @@ import java.util.Objects;
 import static java.util.Objects.nonNull;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
-public class Book {
+public class Book implements PublicationContext {
 
     private String title;
     private Series series;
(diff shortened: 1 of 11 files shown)


src/main/java/no/unit/nva/model/contexttypes/Book.java
```diff
@@ -0,0 +1,163 @@
...
+    private String title;
+    private Series series;
+    private Level level;
+    private boolean openAccess;
+    private boolean peerReviewed;
+    @JsonProperty("isbn")
+    private List<String> isbns;
+
+    public Book() {
+
```
axthosarouris, line 27: extra line

src/main/java/no/unit/nva/model/contexttypes/utils/IssnUtil.java
```diff
@@ -0,0 +1,27 @@
+package no.unit.nva.model.contexttypes.utils;
+
+import no.unit.nva.model.exceptions.InvalidIssnException;
+import no.unit.nva.model.validator.IssnValidator;
+
+import static java.util.Objects.isNull;
+
+public class IssnUtil {
```
axthosarouris, line 8: This should be a final class. I see that Jan has disabled the rule [UtilityClass](https://pmd.github.io/latest/pmd_rules_java_design.html#useutilityclass), I do not know why. But this class cannot be extended.

PR #38, opened 2020-05-25
Np 936 publicationinstance journal review

nraanes:
Added PublicationInstance for JournalReview


src/main/java/no/unit/nva/model/instancetypes/PublicationInstance.java
```diff
@@ -8,7 +8,10 @@
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 @JsonSubTypes({
         @JsonSubTypes.Type(name = "JournalArticle", value = JournalArticle.class),
-        @JsonSubTypes.Type(name = "JournalLetter", value = JournalLetter.class)
+        @JsonSubTypes.Type(name = "JournalLetter", value = JournalLetter.class),
+        @JsonSubTypes.Type(name = "JournalLeader", value = JournalLeader.class),
```
nraanes, line 12: Should JournalLeader be here? I added it as it was missing

brinxmat: I think I have put journal letter in twice, which is unsurprising as as when I think about them, they sound the same…

brinxmat: It should

PR #41, opened 2020-05-26
Np 924 create type degree

brinxmat:
- Add type Degree
- Add builders to types
- Migrate series information to body of types
  - Remove Series
  - Move title to Journal
  - Add seriesTitle to Book
  . Add seriesNumber to Book


src/main/java/no/unit/nva/model/contexttypes/Book.java
```diff
@@ -1,21 +1,16 @@
...
 import java.util.Objects;
 
-import static java.util.Objects.nonNull;
-
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 public class Book implements PublicationContext {
 
-    private String title;
-    private Series series;
+    private String seriesTitle;
```
axthosarouris, line 12: Book has no title? or is a book considered an series of books with only one volume?

src/main/java/no/unit/nva/model/contexttypes/Degree.java
```diff
@@ -0,0 +1,71 @@
...
+        super();
+        setSeriesTitle(builder.seriesTitle);
+        setSeriesNumber(builder.seriesNumber);
+        setLevel(builder.level);
+        setOpenAccess(builder.openAccess);
+        setPeerReviewed(builder.peerReviewed);
+        setIsbnList(builder.isbnList);
+    }
+
+
```
axthosarouris, line 25: Just in case implement an equals and hashcode that call the methods of the parent class

src/main/java/no/unit/nva/model/contexttypes/Degree.java
```diff
@@ -0,0 +1,71 @@
+package no.unit.nva.model.contexttypes;
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import no.unit.nva.model.Level;
+
+import java.util.List;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class Degree extends Book {
```
axthosarouris, line 9: What is a Degree?

src/main/java/no/unit/nva/model/contexttypes/Book.java
brinxmat: Book has a title, but this is in entity description, here we are talking about the container format.

src/main/java/no/unit/nva/model/contexttypes/Degree.java
brinxmat: A thesis in practice, I do not know why this name has been chosen.

src/test/java/no/unit/nva/model/contexttypes/BookTest.java
```diff
@@ -29,11 +26,10 @@ void objectMapperReturnsBookWhenInputIsValidJson() throws JsonProcessingExceptio
         String expectedTitle = "The history of JSON in this project";
         Level expectedLevel = Level.LEVEL_2;
         String expectedIsbn = "123 123123 1112";
-        String json = generateJsonFromTemplate(expectedType, expectedTitle, null, expectedLevel,
+        String json = generateJsonFromTemplate(null, expectedLevel,
                 true, true, expectedIsbn);
         Book book = objectMapper.readValue(json, Book.class);
-        assertEquals(expectedTitle, book.getTitle());
-        assertEquals(NullSeries.class, book.getSeries().getClass());
+        assertNull(book.getSeriesTitle());
```
axthosarouris, line 32: It would be better if you created a book with the same attributes and then checked the equality on book.
`assertEquals(actualBook,expectedBook)

src/main/java/no/unit/nva/model/contexttypes/Degree.java
brinxmat: This ends logically the same as super.equals(object), which is what we are calling

PR #42, opened 2020-05-27
Np 940 publicationinstance report

nraanes:
Added all PublicationInstances for Report.
As of now all four of these are identical. When we get more information we can refactor and/or add stuff as needed.


src/test/java/no/unit/nva/model/instancetypes/ReportPolicyTest.java
```diff
@@ -0,0 +1,52 @@
...
+        String begin = "2";
+        String end = "3";
+        ReportPolicy expected = generateReportPolicy(begin, end);
+
+        String json = ReportContentUtil.generateJsonString(type, begin, end, true);
+        assertEquals(expected, objectMapper.readValue(json, ReportPolicy.class));
+    }
+
+    private ReportPolicy generateReportPolicy(String begin, String end) throws InvalidPageTypeException {
+        Range pages = new Range.Builder()
```
brinxmat, line 43: MonographPages, which affects all of the tests.

src/test/java/no/unit/nva/model/instancetypes/ReportResearchTest.java
```diff
@@ -0,0 +1,52 @@
...
+        String begin = "2";
+        String end = "3";
+        ReportResearch expected = generateReportResearch(begin, end);
+
+        String json = ReportContentUtil.generateJsonString(type, begin, end, true);
+        assertEquals(expected, objectMapper.readValue(json, ReportResearch.class));
+    }
+
+    private ReportResearch generateReportResearch(String begin, String end) throws InvalidPageTypeException {
+        Range pages = new Range.Builder()
```
brinxmat, line 43: MonographPages, which affects all of the tests

src/test/java/no/unit/nva/model/instancetypes/ReportTest.java
```diff
@@ -0,0 +1,52 @@
...
+        String begin = "2";
+        String end = "3";
+        Report expected = generateReport(begin, end);
+
+        String json = ReportContentUtil.generateJsonString(type, begin, end, true);
+        assertEquals(expected, objectMapper.readValue(json, Report.class));
+    }
+
+    private Report generateReport(String begin, String end) throws InvalidPageTypeException {
+        Range pages = new Range.Builder()
```
brinxmat, line 43: MonographPages, which affects all of the tests

src/test/java/no/unit/nva/model/instancetypes/ReportWorkingPaperTest.java
```diff
@@ -0,0 +1,53 @@
...
+        String end = "3";
+        ReportWorkingPaper expected = generateReportWorkingPaper(begin, end);
+
+        String json = ReportContentUtil.generateJsonString(type, begin, end, true);
+        assertEquals(expected, objectMapper.readValue(json, ReportWorkingPaper.class));
+    }
+
+    private ReportWorkingPaper generateReportWorkingPaper(String begin, String end)
+            throws InvalidPageTypeException {
+        Range pages = new Range.Builder()
```
brinxmat, line 44: MonographPages, which affects all of the tests

src/test/java/no/unit/nva/model/util/ReportContentUtil.java
```diff
@@ -0,0 +1,29 @@
+package no.unit.nva.model.util;
+
+public class ReportContentUtil {
```
brinxmat, line 3: ReportContentTestUtil

axthosarouris:
I am not sure about it, but having only one reviewer in a PR, gives me a weird feeling. Maybe too few people are aware or can understand the changes? I guess given the time pressure it is OK

src/main/java/no/unit/nva/model/instancetypes/ReportContent.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public Pages getPages() {
+        return pages;
+    }
+
+    @Override
+    public void setPages(Pages pages) throws InvalidPageTypeException {
+        if (nonNull(pages) && !(pages instanceof Range)) {
```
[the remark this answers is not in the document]
nraanes: Fixed

src/test/java/no/unit/nva/model/instancetypes/ReportPolicyTest.java
nraanes: Hopefully resolved now!

src/main/java/no/unit/nva/model/instancetypes/ReportContent.java
```diff
@@ -0,0 +1,82 @@
...
+    private boolean peerReviewed;
+
+    /**
+     * This constructor ensures that the peerReviewed value is always false.
+     *
+     * @param pages         the Pages of the PublicationInstance.
+     * @param peerReviewed  the value is always ignored.
+     * @throws InvalidPageTypeException if the type of Pages is incompatible with the PublicationInstance type.
+     */
+    @SuppressWarnings("PMD.ConstructorCallsOverridableMethod")
```
axthosarouris, line 30: Somehow it seems that PublicationInstance could be an abstract class. That could maybe simplify the code a bit. There is no class that practically extends another class that is not PublicationInstance and itself is a PublicationInstance. 

If PublicationInstance were an abstract class, then some "anti-patterns"  (although too harsh of a characterization here) could be avoided and some classes could become smaller.  
This is not to be solved here, but should we create an issue for this?

src/main/java/no/unit/nva/model/instancetypes/ReportPolicy.java
```diff
@@ -0,0 +1,42 @@
+package no.unit.nva.model.instancetypes;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import no.unit.nva.model.exceptions.InvalidPageTypeException;
+import no.unit.nva.model.pages.Pages;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class ReportPolicy extends Report {
```
axthosarouris, line 10: is a ReportPolicy and a Report equal if the have the same values in their fields?

src/main/java/no/unit/nva/model/instancetypes/ReportResearch.java
```diff
@@ -0,0 +1,42 @@
+package no.unit.nva.model.instancetypes;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import no.unit.nva.model.exceptions.InvalidPageTypeException;
+import no.unit.nva.model.pages.Pages;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class ReportResearch extends Report {
```
axthosarouris, line 10: Are a ReportResearch and a ReportPolicy if they share the same values in their fields?

src/main/java/no/unit/nva/model/instancetypes/ReportContent.java
```diff
@@ -0,0 +1,82 @@
...
+    public ReportContent(Pages pages, boolean peerReviewed) throws InvalidPageTypeException {
+        super();
+        setPages(pages);
+        if (peerReviewed) {
```
axthosarouris, line 34: wouldn't be simple that isPeerReviewed returns always false and one does not pass the argument in the constructor at all? setPeerReviewed could simple do nothing.

src/main/java/no/unit/nva/model/instancetypes/ReportPolicy.java
nraanes: They are equal as of now, but I believe they will differ when we know more about the types.

src/main/java/no/unit/nva/model/instancetypes/ReportResearch.java
nraanes: They are equal as of now, but I believe they will differ when we know more about the types.

src/test/java/no/unit/nva/model/instancetypes/ReportPolicyTest.java
```diff
@@ -0,0 +1,63 @@
...
+        String type = "ReportPolicy";
+        String pages = "42";
+        String introductionBegin = "1";
+        String introductionEnd = "3";
+        boolean illustrated = false;
+        ReportPolicy expected = generateReportPolicy(pages, introductionBegin, introductionEnd, illustrated);
+
+        String json = ReportContentTestUtil.generateJsonString(type, pages, introductionBegin, introductionEnd,
+                illustrated, false);
+        assertEquals(expected, objectMapper.readValue(json, ReportPolicy.class));
```
axthosarouris, line 43: Two exceptions can occur here. One while trying to read an invalid ReportPolicy.json file and another if the condition fails. Assign the actual value to a variable.

src/test/java/no/unit/nva/model/instancetypes/ReportPolicyTest.java
```diff
@@ -0,0 +1,63 @@
...
+            InvalidPageTypeException {
+        ReportPolicy expected = generateReportPolicy("42", "1", "3", false);
+        String json = objectMapper.writeValueAsString(expected);
+        ReportPolicy reportPolicy = objectMapper.readValue(json, ReportPolicy.class);
+        assertEquals(expected, reportPolicy);
+    }
+
+    @DisplayName("Report policy cannot be peer reviewed")
+    @Test
+    void reportPolicySetsPeerReviewedToFalseWhenPeerReviewIsTrue() throws JsonProcessingException,
```
axthosarouris, line 31: where is peerReview set to true?

src/main/java/no/unit/nva/model/instancetypes/ReportContent.java
```diff
@@ -0,0 +1,82 @@
...
+            logger.warn(SETTING_PEER_REVIEWED_FALSE, ReportContent.class.getSimpleName());
```
axthosarouris, line 35: This warning should be in setPeerReviewed() and the method should not be called here. In fact what you want is 

```
public boolean getPeerReviewed(){
  return false;
}

public boolean setPeerReviewed(){
  logger.warn("You cannot do that");
}
```

src/test/java/no/unit/nva/model/instancetypes/ReportTest.java
```diff
@@ -0,0 +1,63 @@
...
+            InvalidPageTypeException {
+        Report expected = generateReport("42", "1", "3", false);
+        String json = objectMapper.writeValueAsString(expected);
+        Report report = objectMapper.readValue(json, Report.class);
+        assertEquals(expected, report);
+    }
+
+    @DisplayName("Report cannot be peer reviewed")
+    @Test
+    void reportSetsPeerReviewedToFalseWhenPeerReviewIsTrue() throws JsonProcessingException,
```
axthosarouris, line 31: where is peerReview set to true? I fail to see that.

src/test/java/no/unit/nva/model/instancetypes/ReportWorkingPaperTest.java
```diff
@@ -0,0 +1,65 @@
...
+            InvalidPageTypeException {
+        ReportWorkingPaper expected = generateReportWorkingPaper("42", "1", "3", false);
+        String json = objectMapper.writeValueAsString(expected);
+        ReportWorkingPaper reportWorkingPaper = objectMapper.readValue(json, ReportWorkingPaper.class);
+        assertEquals(expected, reportWorkingPaper);
+    }
+
+    @DisplayName("Report working paper cannot be peer reviewed")
+    @Test
+    void reportWorkingPaperSetsPeerReviewedToFalseWhenPeerReviewIsTrue() throws JsonProcessingException,
```
axthosarouris, line 31: Ditto

src/main/java/no/unit/nva/model/instancetypes/ReportContent.java
nraanes: Should I change this @brinxmat ?

brinxmat: c.f. https://github.com/BIBSYSDEV/nva-datamodel-java/blob/develop/src/main/java/no/unit/nva/model/instancetypes/JournalNonPeerReviewedContent.java

nraanes:
> I would recommend that you remove the field peerReviewed from the classes where the field is always false. just make the getter return false. This way the intention is clearer, that peerReviewed is always false.

peerReviewed field and usage in constructors removed

src/main/java/no/unit/nva/model/instancetypes/ReportContent.java
nraanes: I made a task for this :)

src/main/java/no/unit/nva/model/instancetypes/BookAnthology.java
```diff
@@ -1,31 +1,28 @@
...
-import no.unit.nva.model.exceptions.InvalidPageTypeException;
-import no.unit.nva.model.pages.Pages;
+import no.unit.nva.model.pages.MonographPages;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
-public class BookAnthology extends BookMonograph {
-    public BookAnthology() {
-        super();
+public class BookAnthology extends PeerReviewedMonograph {
+    public BookAnthology(@JsonProperty("pages") MonographPages pages,
```
axthosarouris, line 9: add an empty line here?

PR #46, opened 2020-05-30
Np 925 create type chapter

brinxmat:
- Refactor interfaces to allow subtypes with link
- Add Chapter
- Add tests

### DIFF
--- a/src/main/java/no/unit/nva/model/contexttypes/BaseContext.java
+++ b/src/main/java/no/unit/nva/model/contexttypes/BaseContext.java
@@ -1,0 +1,18 @@
+package no.unit.nva.model.contexttypes;
+
+import no.unit.nva.model.Level;
+
+public interface BaseContext extends PublicationContext {
+
+    Level getLevel();
+
+    void setLevel(Level level);
+
+    boolean isOpenAccess();
+
+    void setOpenAccess(boolean openAccess);
+
+    boolean isPeerReviewed();
+
+    void setPeerReviewed(boolean peerReviewed);
+}
--- a/src/main/java/no/unit/nva/model/contexttypes/LinkedContext.java
+++ b/src/main/java/no/unit/nva/model/contexttypes/LinkedContext.java
@@ -2,6 +2,10 @@ package no.unit.nva.model.contexttypes;
 
 import java.net.URI;
 
+/**
+ * LinkedContexts are contexts that are expressed as a URI, allowing an existing (external) Publication
+ * to be used as a PublicationContext. For example, a BookAnthology containing a Chapter.
+ */
 public interface LinkedContext extends PublicationContext {
 
     URI getLinkedContext();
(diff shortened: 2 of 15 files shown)


src/main/java/no/unit/nva/model/Contributor.java
```diff
@@ -8,6 +8,7 @@
 import java.util.List;
 import java.util.Objects;
 import no.unit.nva.model.exceptions.MalformedContributorException;
+import nva.commons.utils.JacocoGenerated;
```
axthosarouris, line 11: Make a PR only with these changes and branch again for other changes.

src/main/java/no/unit/nva/model/contexttypes/BaseContext.java
```diff
@@ -0,0 +1,18 @@
+package no.unit.nva.model.contexttypes;
+
+import no.unit.nva.model.Level;
+
+public interface BaseContext extends PublicationContext {
```
axthosarouris, line 5: What is a `BaseContext`? What is the difference between a `BaseContext` and a `PublicationContext`?  When is something a `PublicationContext` but not a `BaseContext`? If we move all the methods of `BaseContext` to `PublicationContext` is it a logical error? 

I think a Javadoc documentation exlaining what is this class would help a lot the future developer who will join this project.

src/main/java/no/unit/nva/model/contexttypes/LinkedContext.java
```diff
@@ -0,0 +1,10 @@
+package no.unit.nva.model.contexttypes;
+
+import java.net.URI;
+
+public interface LinkedContext extends PublicationContext {
```
axthosarouris, line 5: This kind of explains what is the difference between the `LinkedContext` and the `BaseContext`. Although I kind of understand what is the difference between the `BaseContext`  and what the `LinkedContext`, it not tottaly clear. A documentation text (or link) would be helpful

src/test/java/no/unit/nva/model/contexttypes/ChapterTest.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @DisplayName("Chapter: ObjectMapper throws JsonMappingException when linkingContext URI is null")
+    @ParameterizedTest
+    @NullSource
+    void objectMapperThrowsJsonMappingExceptionWhenInputIsNull(String input) {
+        JsonMappingException exception = assertThrows(JsonMappingException.class,
+            () -> objectMapper.readValue(generateChapterJson(input), Chapter.class));
+
+        assertTrue(exception.getMessage(). contains(String.format(Chapter.ERROR_TEMPLATE, input)));
```
axthosarouris, line 45: 1. put the expectedString as a separate variable. Let the compiler do the optimization for us.
2.  When this test fails the error message will be expected `true` but found `false`.  Suggestion: 
```
build.gradle:
   testImplementation group: 'org.hamcrest', name: 'hamcrest', version: '2.2'

ChapterTest:
String expectedErrorMessage= String.format(Chapter.ERROR_TEMPLATE, input);
assertThat(exception.getMessage(), containsString(expectedErrorMessage));
```

axthosarouris:
The requested change from before was about the unmerged PR.

PR #47, opened 2020-05-31
Add support for books that are monographs

brinxmat:
- Add instanceType BookMonograph
- Update PublicationInstance for subtype
- Add JsonTypeInfo and constructor to MonographPages
- Add tests for BookMonograph
- Update related tests for JournalArticle and Range


src/test/java/no/unit/nva/model/instancetypes/BookMonographTest.java
```diff
@@ -0,0 +1,145 @@
...
+    @Test
+    void bookMonographExists() {
+        new BookMonograph();
+    }
+
+    @DisplayName("BookMonograph: ObjectMapper correctly deserializes object")
+    @Test
+    void objectMapperReturnsBookMonographWhenInputIsValid() throws JsonProcessingException {
+        String expectedIntroductionBegin = "i";
+        String expectedIntroductionEnd = "xxviii";
```
axthosarouris, line 34: LATIN_NUM_1, LATIN_NUM_28

src/test/java/no/unit/nva/model/instancetypes/BookMonographTest.java
```diff
@@ -0,0 +1,145 @@
...
+        String expectedPages = "398";
```
axthosarouris, line 35: It is a bit confusing because when I see all three lines (33-35) together.  Somehow I do in my head  the calculation (xxviii -i) and I expect it to be 398. Why is it important for the introduction pages to be in this test? Can you make another test where you say for example,
`getPagesDoesNotIncludeIntroductoryPages()` or sth like that?

src/main/java/no/unit/nva/model/instancetypes/BookMonograph.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Override
+    public Pages getPages() {
+        return pages;
+    }
+
+    @Override
+    public void setPages(Pages pages) throws InvalidPageTypeException {
+        if (nonNull(pages) && !(pages instanceof MonographPages)) {
```
axthosarouris, line 36: What if we  defined PublicationInstance like this:

```
public interface PublicationInstance<P extends Pages> {

    P getPages();

    void setPages(P  pages) throws InvalidPageTypeException;

    void setPeerReviewed(boolean peerReviewed);

    boolean isPeerReviewed();
```

then we could set pages here to be MonographPages.

src/test/java/no/unit/nva/model/instancetypes/DegreeBachelorTest.java
```diff
@@ -0,0 +1,136 @@
...
+                .build();
+        InvalidPageTypeException exception = assertThrows(InvalidPageTypeException.class, executable);
+        String expectedMessage = generateInvalidPageTypeExceptionMessage(DegreeBachelor.class);
+        assertEquals(expectedMessage, exception.getMessage());
+    }
+
+    @DisplayName("DegreeBachelor does not throw InvalidPageTypeException when input is null")
+    @ParameterizedTest
+    @NullSource
+    void degreeBachelorThrowsInvalidPageTypeExceptionWhenInputIsNull(Pages pages) {
```
axthosarouris, line 112: the method name is wrong probably

src/test/java/no/unit/nva/model/instancetypes/DegreeMasterTest.java
```diff
@@ -0,0 +1,135 @@
...
+                .build();
+        InvalidPageTypeException exception = assertThrows(InvalidPageTypeException.class, executable);
+        String expectedMessage = generateInvalidPageTypeExceptionMessage(DegreeMaster.class);
+        assertEquals(expectedMessage, exception.getMessage());
+    }
+
+    @DisplayName("DegreeMaster does not throw InvalidPageTypeException when input is null")
+    @ParameterizedTest
+    @NullSource
+    void degreeMasterThrowsInvalidPageTypeExceptionWhenInputIsNull(Pages pages) {
```
axthosarouris, line 111: method name.

src/test/java/no/unit/nva/model/instancetypes/DegreePhdTest.java
```diff
@@ -0,0 +1,137 @@
...
+                .build();
+        InvalidPageTypeException exception = assertThrows(InvalidPageTypeException.class, executable);
+        String expectedMessage = generateInvalidPageTypeExceptionMessage(DegreePhd.class);
+        assertEquals(expectedMessage, exception.getMessage());
+    }
+
+    @DisplayName("DegreePhd does not throw InvalidPageTypeException when input is null")
+    @ParameterizedTest
+    @NullSource
+    void degreePhdThrowsInvalidPageTypeExceptionWhenInputIsNull(Pages pages) {
```
axthosarouris, line 113: wrong name

src/main/java/no/unit/nva/model/instancetypes/BookMonograph.java
axthosarouris: But this is just a suggestion and not for now

brinxmat: I think this is an idea worth looking at, and was thinking much the same as I was looking at the tests here and for Journal*. today. I added this to the refactor PublicationInstance task, which I will look at next after merging

PR #54, opened 2020-06-03
Fix range handling

brinxmat:
Given the requested changes in the. BookMonograph branch, it turned out that there was some odd cases that appeared as I wrote. the tests. This PR is the. result of the discoveries from. the QA:

- Ranges can be null
- Ranges cannot be null and a value (in either order)
- Ranges cannot be empty strings
- Range data is sanitized

As a result, we add:

- checks for the. above
- InvalidPageRangeException
- Some minor tweaks to the tests (incorporated from the BookMonograph QA)
- Additional testing for the cases mentioned above

### DIFF
--- a/src/main/java/no/unit/nva/model/pages/Range.java
+++ b/src/main/java/no/unit/nva/model/pages/Range.java
@@ -26,7 +26,8 @@ public class Range implements Pages {
      *
      * @param begin The beginning of the range.
      * @param end   The end of the range
-     * @throws InvalidPageRangeException thrown if one of beginning or end is null.
+     * @throws InvalidPageRangeException thrown if begin is null and end is not, or vice versa,
+     *                                   or if either or both values are empty.
      */
     @JsonCreator
     public Range(@JsonProperty("begin") String begin, @JsonProperty("end") String end) throws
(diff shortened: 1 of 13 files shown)


src/main/java/no/unit/nva/model/pages/Range.java
```diff
@@ -1,39 +1,97 @@
...
+    private final String begin;
+    private final String end;
+
+    /**
+     * Default constructor for Range ensures that a range has a specified beginning and an end, and contains
+     * no unnecessary whitespace, nor members that are only whitespace.
+     *
+     * @param begin The beginning of the range.
+     * @param end   The end of the range
+     * @throws InvalidPageRangeException thrown if one of beginning or end is null.
```
axthosarouris, line 29: rephrase to if either beginning or end is null but not both ?

src/test/java/no/unit/nva/model/pages/RangeTest.java
```diff
@@ -0,0 +1,86 @@
...
+        Executable executable = () -> new Range(begin, end);
+        InvalidPageRangeException exception = assertThrows(InvalidPageRangeException.class, executable);
+        String expectedMessage = String.format(InvalidPageRangeException.ERROR_TEMPLATE,
+                EMPTY_STRING_PLACEHOLDER, EMPTY_STRING_PLACEHOLDER);
+        assertEquals(expectedMessage, exception.getMessage());
+    }
+
+    @DisplayName("Range throws InvalidPageRangeException if only one of the inputs is null")
+    @ParameterizedTest
+    @CsvSource({",x", "i,"})
```
axthosarouris, line 78: Could you add the case of null and whitespace?

PR #57, opened 2020-06-07
Refactoring testing model and peer review handling

brinxmat:
- making sense of the changes that have been made
- removing duplicated code
- unifying test model
- unifying model
- updating tests


src/main/java/no/unit/nva/model/instancetypes/BookAnthology.java
```diff
@@ -1,25 +1,23 @@
...
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import no.unit.nva.model.pages.MonographPages;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 public class BookAnthology extends BookMonograph {
-    public BookAnthology() {
-        super();
+    public BookAnthology(@JsonProperty("pages") MonographPages pages,
```
axthosarouris, line 9: it missed the equals and hash code.

src/main/java/no/unit/nva/model/instancetypes/BookMonograph.java
```diff
@@ -1,96 +1,35 @@
...
-
-import java.util.Objects;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
-public class BookMonograph implements PublicationInstance<MonographPages> {
-
-    private MonographPages pages;
-    private boolean peerReviewed;
-    private boolean openAccess;
+public class BookMonograph extends Monograph {
```
axthosarouris, line 8: equals and hashcode

src/main/java/no/unit/nva/model/instancetypes/Monograph.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof Monograph)) {
+            return false;
+        }
+        Monograph monograph = (Monograph) o;
+        return peerReviewed == monograph.peerReviewed
```
axthosarouris, line 46: This is confusing. I think it is better to use getters in the equals that the actual fields. Some fields are private some others are protected and it is hard to keep an overview on all the fields. On the contrary,  the getters are all public and reveal how the rest of the "world" sees these objects.

src/main/java/no/unit/nva/model/instancetypes/Monograph.java
```diff
@@ -0,0 +1,55 @@
...
+                && Objects.equals(getPages(), monograph.getPages());
```
axthosarouris, line 47: here we use getters, and on the above line the field. I think we should adopt one unique strategy, and that is that all fields are private and if we need to access a field outside a class we use the accessors.

src/main/java/no/unit/nva/model/instancetypes/NonPeerReviewedMonograph.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * In cases where data is inserted with a value peerReviewed = true, an Unexpected exception will be thrown,
+     * otherwise the value is ignored as the class always returns false.
+     *
+     * @param peerReviewed a boolean value.
+     * @throws UnexpectedException thrown if the boolean equals true.
+     */
+    public void setPeerReviewed(boolean peerReviewed) throws UnexpectedException {
+        if (peerReviewed) {
```
axthosarouris, line 40: :100:

src/main/java/no/unit/nva/model/instancetypes/Report.java
```diff
@@ -1,36 +1,38 @@
...
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import no.unit.nva.model.pages.MonographPages;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
-public class Report extends ReportContent {
+public class Report extends NonPeerReviewedMonograph {
 
     /**
-     * Constructor for Report.
+     * The constructor allows setting of pages and open access status, and sets peer-reviewed status to false.
```
axthosarouris, line 11: :100:

PR #61, opened 2020-06-14
Np 1013 prelogue refactoring

brinxmat:
- Added end-to-end round-tripping of all variants of model
- Changed some features in model that were inconsistent
- Removed fields openAccess and Peer reviewed where irrelevant
- Gradually moving away from Publication generator

### DIFF
--- a/src/main/java/no/unit/nva/model/Reference.java
+++ b/src/main/java/no/unit/nva/model/Reference.java
@@ -1,7 +1,7 @@
 package no.unit.nva.model;
 
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
-import no.unit.nva.model.contexttypes.BasicContext;
+import no.unit.nva.model.contexttypes.PublicationContext;
 import no.unit.nva.model.instancetypes.PublicationInstance;
 import nva.commons.utils.JacocoGenerated;
 
@@ -10,9 +10,9 @@ import java.util.Objects;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 public class Reference {
-    private BasicContext publicationContext;
+    private PublicationContext publicationContext;
     private URI doi;
-    private PublicationInstance publicationInstance;
+    private PublicationInstance<?> publicationInstance;
 
     public Reference() {
     }
@@ -23,11 +23,11 @@ public class Reference {
         setPublicationInstance(builder.publicationInstance);
     }
 
-    public BasicContext getPublicationContext() {
+    public PublicationContext getPublicationContext() {
         return publicationContext;
     }
 
-    public void setPublicationContext(BasicContext publicationContext) {
+    public void setPublicationContext(PublicationContext publicationContext) {
         this.publicationContext = publicationContext;
     }
 
@@ -39,20 +39,20 @@ public class Reference {
         this.doi = doi;
     }
 
-    public PublicationInstance getPublicationInstance() {
+    public PublicationInstance<?> getPublicationInstance() {
         return publicationInstance;
     }
 
-    public void setPublicationInstance(PublicationInstance publicationInstance) {
+    public void setPublicationInstance(PublicationInstance<?> publicationInstance) {
         this.publicationInstance = publicationInstance;
     }
 
     public static final class Builder {
-        private PublicationInstance publicationInstance;
-        private BasicContext publicationContext;
+        private PublicationInstance<?> publicationInstance;
+        private PublicationContext publicationContext;
         private URI doi;
 
-        public Builder withPublishingContext(BasicContext publicationContext) {
+        public Builder withPublishingContext(PublicationContext publicationContext) {
             this.publicationContext = publicationContext;
             return this;
         }
@@ -62,7 +62,7 @@ public class Reference {
             return this;
         }
 
-        public Builder withPublicationInstance(PublicationInstance publicationInstance) {
+        public Builder withPublicationInstance(PublicationInstance<?> publicationInstance) {
             this.publicationInstance = publicationInstance;
             return this;
         }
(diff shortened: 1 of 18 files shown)


src/main/java/no/unit/nva/model/instancetypes/JournalLeader.java
```diff
@@ -26,4 +26,38 @@ public JournalLeader(
     ) {
         super(volume, issue, articleNumber, pages, peerReviewed);
     }
+
```
axthosarouris, line 29: needs an equals() override as well

src/test/java/no/unit/nva/PublicationTest.java
```diff
@@ -0,0 +1,111 @@
...
+            "JournalArticle",
+            "JournalLeader",
+            "JournalLetter",
+            "JournalReview",
+            "JournalShortCommunication",
+            "ReportPolicy",
+            "ReportResearch",
+            "ReportWorkingPaper"
+    })
+    void publicationReturnsValidPublicationWhenInputIsValid(String instanceType) throws Exception {
```
axthosarouris, line 40: Are you testing that the type remains same when serializing/deserializing? E.g. a JournalArticle remains a JournalArticle? Or are you testing as well that the field values are preserved? If you are testing the latter then I would suggest using 'nva-testutils' and the following assertions.

```
    testImplementation group: 'com.github.BIBSYSDEV', name:'nva-testutils', version:'0.1.9'
...
        assertThat(publication,doesNotHaveNullOrEmptyFields());
        assertThat(roundTripped,is(equalTo(publication)));

```

src/test/java/no/unit/nva/model/PublicationJournalArticleTest.java
```diff
@@ -40,7 +41,7 @@ public void objectMappingOfPublicationClassReturnsSerializedJsonWithJsonLdFrame(
     }
 
     @DisplayName("Test publications can be serialized/deserialized")
-    @ParameterizedTest
+    @ParameterizedTest(name = "Test Publication context Journal with Instance type {0} can be (de-)serialized")
```
axthosarouris, line 44: In this test you should verify that all fields of either the expected object or the actual object are non empty

src/test/java/no/unit/nva/model/contexttypes/DegreeTest.java
```diff
@@ -74,11 +74,11 @@ void objectMapperReturnsDegreeWhenInputIsValidJson(String seriesTitle,
     @DisplayName("Degree serializes expected json")
     @ParameterizedTest
     @CsvSource({
-            "A series title,123,Full publisher details,LEVEL_2,true,true,\"9780201309515|9788131700075\"",
-            ",, Full publisher details, LEVEL_2, true, true, \"9780201309515|9788131700075\"",
-            "A series title,,Full publisher details,LEVEL_2,true,true,\"9780201309515|9788131700075\"",
-            "Fulong of Oolong,12,T Publishing,LEVEL_0,false,false,\"9780201309515|9788131700075\"",
-            "A Marxist analysis of marking systems,6903,ACO,LEVEL_1,true,false,"
+            "A series title,123,Full publisher details,,true,true,\"9780201309515|9788131700075\"",
```
axthosarouris, line 77: Why do you keep the level parameter, since it is null in all examples?

src/test/java/no/unit/nva/model/contexttypes/DegreeTest.java
```diff
@@ -74,11 +74,11 @@ void objectMapperReturnsDegreeWhenInputIsValidJson(String seriesTitle,
...
+            ",, Full publisher details,, true, true, \"9780201309515|9788131700075\"",
```
axthosarouris, line 78: Although I understand the practicality of the parametrized tests, it is not obvious what is the role of some of the examples. 
I understand that the 1st line contains all information (except for Level), the 2nd line is missing a title, the 3rd is missing the series number, the 5th is missing the ISBNs , but what is the 4th line missing?

Also I am not sure what is this test testing. 
Case A:
We are testing the JSON format i.e. new lines and spaces or is it testing the existence of the fields? If that is the case, why not test the configuration of the JSON mapper instead? Also what are the expectations? Are we expecting the empty fields to be non-existent or are we expecting them to be null? 

Case B:
We are testing the various cases where a Degree instance can be valid. i.e., A degree with all fields is filled is valid, A degree without title is valid, etc. 

Case C:
We are testing that the various valid formats of a degree can be serialized/de-serialized without data loss. 

The values of open access and peer review are never used in the Degree, and they are always set to false by the code (wrong usage of Boolean.getBoolean()),

src/main/java/no/unit/nva/model/instancetypes/JournalLeader.java
```diff
@@ -26,4 +26,38 @@ public JournalLeader(
...
+    public static final class Builder {
```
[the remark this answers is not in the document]
brinxmat: Logically equivalent til super() in both cases

src/main/java/no/unit/nva/model/Reference.java
```diff
@@ -10,9 +10,9 @@
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 public class Reference {
-    private BasicContext publicationContext;
+    private PublicationContext publicationContext;
     private URI doi;
-    private PublicationInstance publicationInstance;
+    private PublicationInstance<?> publicationInstance;
```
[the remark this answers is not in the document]
axthosarouris: Doesn't this mean that the type has to be declared with the class? i.e. the `Reference` class to become `Reference<DESCRIPTIVE_NAME>`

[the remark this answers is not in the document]
janbj: Maybe, but I think it's also possible within a method. Anywho, it's not very important in this case.

PR #64, opened 2020-06-20
Np 1013 restucture package membership

brinxmat:
- move instance types to named packages
- simplify class inheritance
- refactor tests to simplify
- add missing tests


src/main/java/no/unit/nva/model/instancetypes/NonPeerReviewed.java
```diff
@@ -0,0 +1,20 @@
+package no.unit.nva.model.instancetypes;
+
+import java.rmi.UnexpectedException;
+
+public class NonPeerReviewed {
```
axthosarouris, line 5: Maybe it should be abstract

src/main/java/no/unit/nva/model/instancetypes/NonPeerReviewed.java
```diff
@@ -0,0 +1,20 @@
...
+    public static final String PEER_REVIEWED_ERROR_TEMPLATE = "%s is assumed not to be peer-reviewed";
+
+    /**
+     * In cases where data is inserted with a value peerReviewed = true, an Unexpected exception will be thrown,
+     * otherwise the value is ignored as the class always returns false.
+     *
+     * @param peerReviewed a boolean value.
+     * @throws UnexpectedException thrown if the boolean equals true.
+     */
+    public void setPeerReviewed(boolean peerReviewed) throws UnexpectedException {
```
axthosarouris, line 15: make it final. Add also `getPeerReviewed()` returning false and make it final too

src/main/java/no/unit/nva/model/instancetypes/NonPeerReviewed.java
axthosarouris:
```suggestion

public abstract class NonPeerReviewed<P extends Pages>  implements PublicationInstance<P>{
    public static final String PEER_REVIEWED_ERROR_TEMPLATE = "%s is assumed not to be peer-reviewed";

    /**
     * In cases where data is inserted with a value peerReviewed = true, an Unexpected exception will be thrown,
     * otherwise the value is ignored as the class always returns false.
     *
     * @param peerReviewed a boolean value.
     * @throws UnexpectedException thrown if the boolean equals true.
     */
    public final void setPeerReviewed(boolean peerReviewed) throws UnexpectedException {
        if (peerReviewed) {
            throw new UnexpectedException(String.format(PEER_REVIEWED_ERROR_TEMPLATE, this.getClass().getSimpleName()));
        }
    }

    @Override
    public final boolean isPeerReviewed() {
        return false;
    }
}

```

src/main/java/no/unit/nva/model/instancetypes/report/Report.java
```diff
@@ -1,18 +1,18 @@
-package no.unit.nva.model.instancetypes;
+package no.unit.nva.model.instancetypes.report;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import no.unit.nva.model.instancetypes.NonPeerReviewedMonograph;
 import no.unit.nva.model.pages.MonographPages;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 public class Report extends NonPeerReviewedMonograph {
```
axthosarouris, line 8: There are two Report classes

src/test/java/no/unit/nva/model/instancetypes/TestData.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private final String begin;
+    private final String end;
+    private boolean peerReviewed;
+
+    /**
+     * Creates a random set of data for the Range that is present in each Pages subtype.
+     */
+    public TestData() {
+        begin = random(1, 12);
```
axthosarouris, line 19: I do not understand the necessity of random data. Is there a case where a specific combination might throw an Exception? Then this combination should be tested.  Is it to ensure that values from other tests are not surviving and some tests are passing due to surviving values between tests? This could mean that we should simplify our tests.

PR #65, opened 2020-06-25
Np 1013 restucture package membership fix tests

brinxmat:
- Fix testing of JSON
- Add PeerReviewed class to mirror NonPeerReviewed

### DIFF
--- a/src/main/java/no/unit/nva/model/instancetypes/NonPeerReviewed.java
+++ b/src/main/java/no/unit/nva/model/instancetypes/NonPeerReviewed.java
@@ -14,7 +14,7 @@ public abstract class NonPeerReviewed<P extends Pages> implements PublicationIns
      * @param peerReviewed a boolean value.
      * @throws UnexpectedException thrown if the boolean equals true.
      */
-    public void setPeerReviewed(boolean peerReviewed) throws UnexpectedException {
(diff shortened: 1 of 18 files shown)


src/main/java/no/unit/nva/model/instancetypes/NonPeerReviewed.java
```diff
@@ -12,9 +14,14 @@
      * @param peerReviewed a boolean value.
      * @throws UnexpectedException thrown if the boolean equals true.
      */
-    public void setPeerReviewed(boolean peerReviewed) throws UnexpectedException {
+    public final void setPeerReviewed(boolean peerReviewed) throws UnexpectedException {
```
janbj, line 17: This is not actually setting the field even if it is false?

brinxmat: I think that this made sense at one point, I don't think it does following refactoring. Will fix.

axthosarouris: There is no field to set. `isPeerReviewed()` returns always false, so there is no reason for a "reviewed" field to exist. I think that the exception has the meaning that the input data are invalid

PR #67, opened 2020-07-09
Add peer-reviewed test to NonPeerReviewedTest, delete related tests

brinxmat:
The tests should be carried out in one place, so I moved the test and deleted the extra testing


src/test/java/no/unit/nva/model/instancetypes/NonPeerReviewedTest.java
```diff
@@ -0,0 +1,29 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class NonPeerReviewedTest {
+
+    @SuppressWarnings("ConstantConditions")
+    @Test
+    public void nonPeerReviewedMonographPagesSettingPeerReviewedTrueThrowsUnexpectedException() {
+        NonPeerReviewed<MonographPages> nonPeerReviewed = new NonPeerReviewedMonograph();
+        Executable executable = () -> nonPeerReviewed.setPeerReviewed(true);
+        assertThrows(UnexpectedException.class, executable);
```
axthosarouris, line 19: I hate to do that but I was curious what the UnexpectedException was :) I suggest we can leave it for now, but maybe IllegalArgumentException would be more appropriate?

From the documentation of UnexpectedException:
```
/**
 * An <code>UnexpectedException</code> is thrown if the client of a
 * remote method call receives, as a result of the call, a checked
 * exception that is not among the checked exception types declared in the
 * <code>throws</code> clause of the method in the remote interface.
 *
 * @author  Roger Riggs
 * @since   1.1
 */
```

axthosarouris: We can create a task for this, or just use the IllegalArgumentException if it does not cause many problems.

PR #72, opened 2020-07-31
Added doiRequested flag to PublicationResponse

janbj:
Hm, suggestions for a better solution?


src/main/java/no/unit/nva/WithFlags.java
```diff
@@ -14,6 +14,6 @@
      *
      * @param doiRequested  doiRequested
      */
-    void setDoiRequest(Boolean doiRequested);
+    void setDoiRequested(Boolean doiRequested);
```
janbj, line 17: Renamed wrong setter method name in interface

PR #77, opened 2020-09-10
A Doi Request is optional

norrs:
As the Doi Request is optional, ie it has a state where it could not have been created yet, ensure we declare it in the model as well so the caller knows.

### DIFF
--- a/src/main/java/no/unit/nva/WithInternal.java
+++ b/src/main/java/no/unit/nva/WithInternal.java
@@ -2,6 +2,7 @@ package no.unit.nva;
 
 import java.net.URI;
 import java.time.Instant;
+import java.util.Optional;
 import java.util.UUID;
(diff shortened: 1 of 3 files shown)


axthosarouris:
https://stackoverflow.com/questions/26327957/should-java-8-getters-return-optional-type/26328555#26328555

brinxmat (commented):
My initial intuition was that getting DoiRequest should return a List object, empty in the case there are no requests, however, this contradicts the product owner's conceptualisation of how requesting DOIs works. 

I am against using Optional in this way, preferring perhaps a null-object pattern as null actually means something for the process of requesting a DOI here.

While, I'm aware that this is a Draft PR:
- commit complete, running code
- test the code

In this way, we avoid a long process of attempting to understand what you're trying to achieve.

PR #78, opened 2020-09-11
NP-1046: Support for updating doi request status

norrs:


axthosarouris (changes requested):
As @norrs pointed out, `Publication` is both a Data Transfer Object and a Data Access Object. This by itself can be problematic. I am a bit sceptical of inserting business logic functionality in addition directly to this class. because then the class becomes to complicated. 

I would suggest either, a BusinessObject that takes contains the DTO as field or Businness Object that contains no fields and it contains only the logic for the publication. In this way the Business Object manipulates its own data (according to the Object Oriented paradigm)  and the DTO has nothing to do about it. Moreover, the conversion from the DTO to the BO will always be easy and without data loss since the BO does not contain any fields.  For me it seems like a good solution as it combines both what @norrs  is (very correctly) pushing for, and it also leaves the DTOs untouched. Of course the latter is just a suggestion as there can be obvious disadvantages that I cannot see :)

norrs:
Regarding business objects and DTOs, I believe we should solve this in a separate tech debt task, as we want to do this across all modules. We just need to agree about it, if not the scope of NP-1046 will explode. And with this code we avoid data jealously, so its easier to refactor. 

I also suggest that the "DTOs" from `nva-datamodel-java` gets suffixed with Dto's, so the business objects can own the name. (publication, etc), as those are the objects we would like to work with.

What do you think about this @axthosarouris @janbj @brinxmat

norrs:
@axthosarouris I believe I fixed all your feedback

src/test/java/no/unit/nva/model/PublicationTest.java
```diff
@@ -31,4 +43,49 @@ protected Object produceFramedPublication(JsonNode publicationWithContext) throw
...
+    private Publication getPublicationWithoutDoiRequest() throws InvalidIssnException, MalformedContributorException {
+        return PublicationGenerator.generateJournalArticlePublication().copy()
+            .withDoiRequest(null).build();
+    }
+
+    private Publication generatePublicationWithRejectedDoiRequestStatus()
+        throws InvalidIssnException, MalformedContributorException {
+        var publication = PublicationGenerator.generateJournalArticlePublication();
+        var doiRequest = publication.getDoiRequest();
+        return publication.copy().withDoiRequest(doiRequest.copy().withStatus(REJECTED).build()).build();
```
axthosarouris, line 89:
```
var newDoiRequest = publication.getDoiRequest().copy().withStatus().build();
return publication.copy().withDoiRequest(newDoiRequest).build()

File: src/main/java/no/unit/nva/model/Publication.java [hunk 2]
```diff
@@ -105,6 +105,26 @@ public void setDoiRequest(DoiRequest doiRequest) {
...
+     * @throws IllegalArgumentException Invalid status to change to.
+     * @throws IllegalStateException No DoiRequest exists.
+     * @see DoiRequestStatus
+     */
+    public void updateDoiRequestStatus(DoiRequestStatus requestedStatusChange) {
+        if (Objects.isNull(doiRequest)) {
+            throw new IllegalStateException(
+                ERROR_MESSAGE_UPDATEDOIREQUEST_MISSING_DOIREQUEST);
+        }
+        var ourDoiRequest = getDoiRequest();
```
axthosarouris, line 120: ourDoiRequest is the same as this.doiRequest. it is the same object as `this.doiRequest` i.e., `ourDoiRequest` and `this.doiRequest` point to exactly the same object. There is no point in having it. use `this.doiRequest`

src/test/java/no/unit/nva/model/DoiRequestStatusTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+        "REJECTED,REJECTED",
+        "REJECTED,REQUESTED"
+    })
+    void invalidStatusChanges(DoiRequestStatus existingState,
+                              DoiRequestStatus requestedChange) {
+        var actualException = assertThrows(IllegalArgumentException.class,
+            () -> existingState.changeStatus(requestedChange));
+        assertThat(actualException.getMessage(), is(equalTo(
+                String.format(ERROR_MESSAGE_NOT_ALLOWED_TO_CHANGE_STATUS_FROM_S_TO_S,
```
axthosarouris, line 46: create a public static method to avoid replicating code

src/test/java/no/unit/nva/model/PublicationTest.java
```diff
@@ -31,4 +43,49 @@ protected Object produceFramedPublication(JsonNode publicationWithContext) throw
         options.setPruneBlankNodeIdentifiers(true);
         return JsonLdProcessor.frame(input, frame, options);
     }
+
+    @Test
+    public void updatingDoiRequestStatusForValidTransitionChangeReturnsUpdatedDoiRequestStatus()
+        throws InvalidIssnException, MalformedContributorException {
+        var publication = PublicationGenerator.generateJournalArticlePublication();
+        publication.updateDoiRequestStatus(REJECTED);
```
axthosarouris, line 51: parametrize the test.

src/test/java/no/unit/nva/model/PublicationTest.java
```diff
@@ -31,4 +43,49 @@ protected Object produceFramedPublication(JsonNode publicationWithContext) throw
...
+
+    @Test
+    public void updatingDoiRequestStatusWithInvalidTransitionRequestThenThrowsIllegalArgumentException()
+        throws InvalidIssnException, MalformedContributorException {
+        var publication = generatePublicationWithRejectedDoiRequestStatus();
+
+        var actualException = assertThrows(IllegalArgumentException.class,
+            () -> publication.updateDoiRequestStatus(REQUESTED));
+        assertThat(actualException.getMessage(),
+            is(equalTo(String.format(ERROR_MESSAGE_NOT_ALLOWED_TO_CHANGE_STATUS_FROM_S_TO_S, REJECTED, REQUESTED))));
```
axthosarouris, line 66: use public static method for the error message

src/main/java/no/unit/nva/model/DoiRequestStatus.java
```diff
@@ -1,7 +1,46 @@
...
 
+import static java.util.Collections.emptySet;
+
+import java.util.Set;
+
 public enum DoiRequestStatus {
     REQUESTED,
     APPROVED,
     REJECTED;
+    protected static final Set<DoiRequestStatus> validStatusChangeForRecejted = Set.of(APPROVED);
```
axthosarouris, line 11: typo in in Rejected

axthosarouris (changes requested):

src/test/java/no/unit/nva/model/DoiRequestStatusTest.java
```diff
@@ -0,0 +1,48 @@
...
+    void validStatusChanges(DoiRequestStatus existingState,
+                            DoiRequestStatus requestedChange,
+                            DoiRequestStatus expectedState) {
+        assertThat(existingState.changeStatus(requestedChange), is(equalTo(expectedState)));
+    }
+
+    @ParameterizedTest
+    @CsvSource({
+        "REQUESTED,REQUESTED",
+
```
axthosarouris, line 33: remove empty line

src/test/java/no/unit/nva/model/DoiRequestStatusTest.java
```diff
@@ -0,0 +1,48 @@
...
+        "APPROVED,REQUESTED",
+        "APPROVED,APPROVED",
+        "APPROVED,REJECTED",
+
```
axthosarouris, line 37: remove empty line

src/test/java/no/unit/nva/model/PublicationTest.java
axthosarouris: The code looks like you are testing for separate cases. I did not understand that you were testing for REJECTED->APPROVED. Can you add a comment on each line?

```
publication.updateDoiRequestStatus(REJECTED); //initial state
        assertThat(publication.getDoiRequest().getStatus(), is(equalTo(REJECTED)));

        publication.updateDoiRequestStatus(APPROVED); // next state
        assertThat(publication.getDoiRequest().getStatus(), is(equalTo(APPROVED)));
```

src/test/java/no/unit/nva/model/PublicationTest.java
axthosarouris: What are you trying to assert here? That the message is the same as the one in the Publication class? that the message contains both REJECTED and REQUESTED statuses? That the message carries some kind of information? 

If we change the message and we omit for example the new status, then the person who will make the change will compare the line you have written with the line in the code and they think, they are the same so I change the test here and I am OK. But if you explicitly test for the presence of the two statuses then you have made your expectation clear. 

My point is that my understanding of this line is that you expect the message specified in the Publication class to appear to the test. So it is not directly linked to the class but it is indirectly because copying the returned value means 'I expect the specific value'

axthosarouris: How about testing if bot REJECTED and REQUESTED statuses are contained in the message?

norrs:
@axthosarouris Does https://github.com/BIBSYSDEV/nva-datamodel-java/pull/78/commits/9eaf34a3ab8e23664efbeb119eb577d1c4e08143 and https://github.com/BIBSYSDEV/nva-datamodel-java/pull/78/commits/d67d079f0035de9bde322c33e46c51c6212c5d4f help with what it is testing?

norrs:
And https://github.com/BIBSYSDEV/nva-datamodel-java/pull/78/commits/1771a5f94b99dde02f52bff52dfea74fbcfdc0ed :)

src/test/java/no/unit/nva/model/PublicationTest.java
```diff
@@ -60,14 +60,11 @@ public void updatingDoiStatusSuccessfullyChangesToValidNewDoiStatus()
...
-    public void updatingDoiRequestStatusWithInvalidTransitionRequestThenThrowsIllegalArgumentException()
+    public void updatingDoiRequestWithInvalidRequestedStatusChangeThenThrowsIllegalArgumentException()
         throws InvalidIssnException, MalformedContributorException {
         var publication = generatePublicationWithRejectedDoiRequestStatus();
 
-        var actualException = assertThrows(IllegalArgumentException.class,
-            () -> publication.updateDoiRequestStatus(REQUESTED));
-        assertThat(actualException.getMessage(),
-            is(equalTo(String.format(ERROR_MESSAGE_NOT_ALLOWED_TO_CHANGE_STATUS_FROM_S_TO_S, REJECTED, REQUESTED))));
+        assertThrows(IllegalArgumentException.class, () -> publication.updateDoiRequestStatus(REQUESTED));
```
axthosarouris, line 67: Maybe I did not explain it correctly. When we throw an Exception we should also verify that there is some kind of information sent to the caller with the Exception. It has happened in the past that a test was passing because it was expecting a Runtime exception and it passed but the thrown Exception was not thrown where we were expecting to be thrown. That is why we decided to test the expected message as well. 

Another thing is that when we throw an Exception we have some expectations from the message. For example here the initial and the desired status. This gives an insight to the user why their request failed. If i hardcode a message then whenever I correct the message (typo, different expressions, additional information etc.) i will have to correct it multiple tests which are not necessarily insterested in those new additions.

axthosarouris (approved):

axthosarouris (approved):

axthosarouris (approved):

src/main/java/no/unit/nva/model/Publication.java
```diff
@@ -105,6 +105,25 @@ public void setDoiRequest(DoiRequest doiRequest) {
...
 
+    /**
+     * Update a publication with the requested status change.
+     * @param requestedStatusChange Requested status change.
+     * @throws IllegalArgumentException Invalid status to change to.
+     * @throws IllegalStateException No DoiRequest exists.
+     * @see DoiRequestStatus
+     */
+    public void updateDoiRequestStatus(DoiRequestStatus requestedStatusChange) {
+        if (Objects.isNull(doiRequest)) {
```
brinxmat, line 116: Static import would be prettier, I'd have modelled a NullObject, but that's me. I'd also probably have ended up with a mess of classes and some weird interface, so maybe just as well I didn't.

src/test/java/no/unit/nva/model/DoiRequestStatusTest.java
```diff
@@ -0,0 +1,45 @@
...
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+public class DoiRequestStatusTest {
+
+    @ParameterizedTest
```
brinxmat, line 14: Ver explanatory, excellent.

brinxmat (approved):

src/main/java/no/unit/nva/model/Publication.java
norrs: I'll fix it in develop after merge. Thanks for the observation.

PR #81, opened 2020-09-18
NP-1500: DoiRequest messages list not null

axthosarouris:
Bug: The message list in a DoiRequest object is not initialized and therefore is null. SInce it is a list the default should be to empty

### DIFF
--- a/src/main/java/no/unit/nva/model/DoiRequest.java
+++ b/src/main/java/no/unit/nva/model/DoiRequest.java
@@ -1,14 +1,12 @@
 package no.unit.nva.model;
 
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
-import nva.commons.utils.JacocoGenerated;
-
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
-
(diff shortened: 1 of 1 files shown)


brinxmat (approved):

src/main/java/no/unit/nva/model/DoiRequest.java
```diff
@@ -115,7 +115,7 @@ public Builder addMessage(DoiRequestMessage message) {
         }
 
         private void initializeMessagesIfEmpty() {
```
norrs, line 117: Remove this, since you always initialize it in the constructor.

norrs (changes requested):

src/main/java/no/unit/nva/model/DoiRequest.java
axthosarouris: empyList is immutable. I could initialize it as a new ArrayList if this is preferable.

axthosarouris: emptyList() creates an immutable list, so I have to create a new List for adding a message. The alternative would be to initialize the mesages list as a new ArrayList.

norrs (approved):

PR #82, opened 2020-09-23
Add field for textbook content

brinxmat:
Added field for textbook to ChapterArticle and PeerReviewedMonograph

Sarah: se gjerne på documentation/{Types where we expect the textbookContent property}

### DIFF
--- a/documentation/BookAnthology.json
+++ b/documentation/BookAnthology.json
@@ -1,6 +1,6 @@
 {
   "type" : "Publication",
-  "identifier" : "5a0d2a7d-d0f3-4d3c-918f-067d6a37eef7",
+  "identifier" : "955fab17-542d-44fa-a6d0-a2100d8b9ef3",
   "status" : "Published",
   "owner" : "me@example.org",
   "publisher" : {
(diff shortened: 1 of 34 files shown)


documentation/BookAnthology.json
```diff
@@ -1,6 +1,6 @@
 {
   "type" : "Publication",
-  "identifier" : "5a0d2a7d-d0f3-4d3c-918f-067d6a37eef7",
+  "identifier" : "955fab17-542d-44fa-a6d0-a2100d8b9ef3",
```
janbj, line 3: Do we need id as part of the ontology for Publication or is that implicit?

janbj (approved):

documentation/DegreePhd.json
```diff
@@ -10,16 +10,17 @@
...
+  "publishedDate" : "2020-09-23T09:40:34.473624Z",
+  "indexedDate" : "2020-09-23T09:40:34.473605Z",
   "handle" : "https://example.org/fakeHandle/13213",
   "doi" : "https://example.org/yet/another/fake/doi/1231/12311",
   "doiRequest" : {
     "type" : "DoiRequest",
     "status" : "APPROVED",
-    "date" : "2020-09-18T07:35:55.134450Z"
+    "date" : "2020-09-23T09:40:34.473494Z",
+    "messages" : [ ]
```
sarahse, line 23: hva er dette feltet for?

sarahse: nvm så ikke at det var inni doiRequests-objektet

sarahse (approved):

documentation/BookAnthology.json
brinxmat: In essence, the nature of the identifier is not important, however, we have platform expectations, so it becomes important. I will create a code example as a suggested fix for this.

norrs (approved):

norrs:
Only confusing with this pull review is maybe actually finding what is the new changes (as we get diffs with all timestamps and random generated identifiers).  The diff is really a lot smaller :p

PR #83, opened 2020-09-24
Make dates and UUIDs in documentation/*.json files static

brinxmat:
Make dates and UUIDs in generated documents static, removing the unnecessary changes in these files because the dates and UUIDs are autogenerated for each test pass.


janbj (approved):

norrs (approved):
Superhero fix! Thanks!

PR #86, opened 2020-10-13
Renamed doiRequest.date to doiRequest.createdDate

janbj:


axthosarouris (dismissed):

norrs (approved):

src/main/java/no/unit/nva/model/DoiRequest.java
```diff
@@ -12,7 +12,7 @@
 public class DoiRequest {
 
     private DoiRequestStatus status;
-    private Instant date;
+    private Instant createdDate;
```
brinxmat, line 15: If we add `@JsonAlias("date")`, we avoid breaking the API

brinxmat (approved):

norrs (changes requested):
Agree with  @brinxmat , fix the JsonAlias :+1:

brinxmat (commented):
Accidentally clicked Approve, see change request

brinxmat (changes requested):
Did it again

norrs (approved):

axthosarouris (approved):

PR #87, opened 2020-10-16
Add journal corrigendum

brinxmat:
Adds the Instance type Journal corrigendum, I don't think it adds anything new at all, just a new named type

### DIFF
--- a/src/main/java/no/unit/nva/model/instancetypes/journal/JournalCorrigendum.java
+++ b/src/main/java/no/unit/nva/model/instancetypes/journal/JournalCorrigendum.java
@@ -4,6 +4,13 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import no.unit.nva.model.pages.Range;
 
+/**
+ * A Journal corrigendum is a correction to a previously published Journal article/content.
+ *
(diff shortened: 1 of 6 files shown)


src/main/java/no/unit/nva/model/instancetypes/journal/JournalCorrigendum.java
```diff
@@ -0,0 +1,62 @@
+package no.unit.nva.model.instancetypes.journal;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import no.unit.nva.model.pages.Range;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class JournalCorrigendum extends JournalNonPeerReviewedContent {
```
axthosarouris, line 8: Is this an example of a Journal Corrigendum?  (https://doi.org/10.1038/nature11848).

Is it worth putting a short explanation in the Javadoc? Something like 
"Publication correcting another publication published by the same authors. It has its own DOI"?

axthosarouris (dismissed):

axthosarouris (dismissed):

janbj (dismissed):

janbj (approved):

axthosarouris (approved):

PR #92, opened 2020-10-19
Np 1698 doi request modified date

janbj:
New property doiRequest.modifiedDate, set to same value as publication.modifiedDate upon doiRequest status changes,

### DIFF
--- a/src/main/java/no/unit/nva/model/Publication.java
+++ b/src/main/java/no/unit/nva/model/Publication.java
@@ -121,10 +121,13 @@ public class Publication
                 ERROR_MESSAGE_UPDATEDOIREQUEST_MISSING_DOIREQUEST);
         }
 
-        setDoiRequest(getDoiRequest().copy()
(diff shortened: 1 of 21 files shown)


axthosarouris (approved):

norrs (approved):

src/main/java/no/unit/nva/model/Publication.java
```diff
@@ -121,10 +121,13 @@ public void updateDoiRequestStatus(DoiRequestStatus requestedStatusChange) {
                 ERROR_MESSAGE_UPDATEDOIREQUEST_MISSING_DOIREQUEST);
         }
 
-        setDoiRequest(getDoiRequest().copy()
+        Instant now = Instant.now();
+        DoiRequest updatedDoiRequest = getDoiRequest().copy()
```
norrs, line 125: Allowed to use the `var updateDoiRequest =`  and `var now = Instant.now();` if you want.

PR #103, opened 2020-12-16
Add github action and remove travis

brinxmat:


.github/gradle.yml
```diff
@@ -0,0 +1,44 @@
+# This workflow will build a Java project with Gradle
```
axthosarouris, line 1: put it in ./gtihub/workflows/

axthosarouris, line 1: put it in ./gtihub/workflows/

axthosarouris (changes requested):

axthosarouris (approved):

PR #117, opened 2021-02-04
NP-1979: update schema.yaml

axthosarouris:
Updated schema, after running gradle build. No code was written, I only run gradle build. The interesting file is template.yml.


documentation/BookAnthology.json
```diff
@@ -131,8 +131,5 @@
       "approvalStatus" : "APPLIED",
       "applicationCode" : "123123"
     } ]
-  } ],
-  "publisherId" : "http://example.org/org/123",
```
axthosarouris, line 135: These fields were used for indexing in the database and we do not need them anymore

brinxmat (approved):

PR #121, opened 2021-03-02
Publication contains datamodel version

axthosarouris:


src/main/java/no/unit/nva/model/Publication.java
```diff
@@ -21,6 +22,8 @@
 public class Publication
         implements WithIdentifier, WithInternal, WithFile, WithMetadata, WithCopy<Publication.Builder> {
 
+    public static final String MODEL_VERSION= "0.11.8";
```
torarnet, line 25: why is this public? it has a getter

src/test/java/no/unit/nva/model/PublicationTest.java
```diff
@@ -29,6 +30,7 @@
     public static final String PUBLICATION_FRAME_JSON = "src/main/resources/publicationFrame.json";
 
     protected static final ObjectMapper objectMapper = nva.commons.core.JsonUtils.objectMapper;
+    public static final String DATAMODEL_VERSION = "DATAMODEL_VERSION";
```
torarnet, line 33: why is this public. Is it used outside the test class?

torarnet, line 33: Explicit define that it is name for a key or is that not necessary?

torarnet (dismissed):

src/main/java/no/unit/nva/model/Publication.java
```diff
@@ -237,6 +241,16 @@ public void setProjects(List<ResearchProject> projects) {
...
     }
 
+    @JsonProperty("modelVersion")
+    public String getModelVersion(){
+        return MODEL_VERSION;
+    }
+
+    @JsonProperty("modelVersion")
+    public void setModelVersion(){
+        //
```
brinxmat, line 251: State explicit NO-OP if that's your intention

brinxmat (changes requested):

src/test/java/no/unit/nva/model/PublicationTest.java
```diff
@@ -72,6 +74,14 @@ public void updatingDoiRequestStatusWithoutDoiRequestThrowsIllegalStateException
             is(equalTo(Publication.ERROR_MESSAGE_UPDATEDOIREQUEST_MISSING_DOIREQUEST)));
     }
 
+    @Test
+    public void getModelVersionReturnsModelVersionDefinedByGradle()
```
svenngjeto, line 78: Tests can be/is package private in Junit5

janbj (dismissed):

src/test/java/no/unit/nva/model/PublicationTest.java
axthosarouris: This class has some protected methods and package private tests will mess the order of the functions (tests first, helper methods afterwards)

janbj (dismissed):

brinxmat (approved):

PR #125, opened 2021-03-10
Publish own artifact to jitpack

axthosarouris:
Originally wanted to fix wrong Java version choice in Jitpack, but ended up building the artifacts in a controlled and repricable way and publishing them in Jitpack. 

Publishing to Jitpack is not meant as a replacement for Maven Central. It is meant to use during development

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,6 +6,7 @@ plugins {
     id 'nebula.lint' version '16.8.0'
     id 'de.fuerstenau.buildconfig' version '1.1.8'
     id 'idea'
+    id 'maven-publish'
 }
 
 apply plugin: 'de.fuerstenau.buildconfig'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -112,3 +119,41 @@ buildConfig {
...
+                    fromResolutionResult()
+                }
+            }
+            pom {
+                name = project.name
+                description = 'Java implementation of NVA Datamodel'
+                url = 'https://github.com/BIBSYSDEV/nva-datamodel-java'
+                licenses {
+                    license {
+                        name = 'The Apache License, Version 2.0'
```
janbj, line 142: Is this the right license?

mattiahj:
Why not use the sonatype snapshot repository? I don't think we should add a second source for our artifacts.

brinxmat (approved):

janbj (approved):

build.gradle
```diff
@@ -6,12 +6,14 @@ plugins {
     id 'nebula.lint' version '16.8.0'
```
mattiahj, line 6: Why not use the sonatype snapshot repository? I don't think we should add a second source for our artifacts.

mattiahj (dismissed):
Why not use the sonatype snapshot repository? I don't think we should add a second source for our artifacts, and a second way of publishing them.

axthosarouris:
The snapshot would require merging something to main. We want to use Jitpack when we are making changes to two repos at the same time and we want to use the git hash value as a version number. It is a basically a tool for using versions that are not mature for publishing or even merging to main. This allows to test if things build remotely, e.g. the PR is green or your code runs as expected in the Cloud.

I agree that it is a bit of a hassle to have two artifactories, but the aim of having Jitpack is to be able to test some code that exists in one module in another module.  If there is another solution that involves only maven central that would be awesome.

mattiahj:
> The snapshot would require merging something to main. 

I don't think so? I mean, we could have a separate Github Action for snapshots, or use a single one for publishing both releases and snapshots. SNAPSHOTS are by definition usually built from stuff merged to develop.

PR #134, opened 2021-04-20
Add additional identifiers for publications

axthosarouris:


src/main/java/no/unit/nva/model/AdditionalIdentifier.java
```diff
@@ -0,0 +1,65 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import java.util.Objects;
+import nva.commons.core.JacocoGenerated;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class AdditionalIdentifier {
+
+    @JsonProperty("source")
+    private String source;
```
brinxmat, line 12: In a sense, the source is the type, I guess

src/main/java/no/unit/nva/model/AdditionalIdentifier.java
```diff
@@ -0,0 +1,65 @@
...
+    @JsonProperty("identifier")
+    private String identifier;
```
brinxmat, line 14: value?

brinxmat (dismissed):

src/main/java/no/unit/nva/model/AdditionalIdentifier.java
axthosarouris: the type word is reserved for the objectType

brinxmat (approved):

PR #165, opened 2021-07-07
Np 2755 update publication context with url

svenngjeto:
Added linkedContext to Journal, Book, Report and Degree.

### DIFF
--- a/src/main/java/no/unit/nva/model/contexttypes/Book.java
+++ b/src/main/java/no/unit/nva/model/contexttypes/Book.java
@@ -42,7 +42,7 @@ public class Book implements BasicContext, LinkedContext {
         setPeerReviewed(builder.peerReviewed);
         setUrl(builder.url);
         setIsbnList(builder.isbnList);
-        setLinkedContext(builder.linkedContext);
+        setLinkedContextURI(builder.linkedContext);
     }
 
     @Override
@@ -137,6 +137,21 @@ public class Book implements BasicContext, LinkedContext {
         this.publisher = publisher;
     }
 
+    @Override
+    public URI getLinkedContext() {
+        return linkedContext;
+    }
+
+    @Override
+    public void setLinkedContext(String linkedContext) {
+        setLinkedContextURI(URI.create(linkedContext));
+    }
+
+    protected void setLinkedContextURI(URI linkedContext) {
+        this.linkedContext = linkedContext;
+    }
+
+
     @JacocoGenerated
     @Override
     public boolean equals(Object o) {
@@ -172,20 +187,6 @@ public class Book implements BasicContext, LinkedContext {
             getIsbnList(),
             getLinkedContext()
         );
-    }
-
-    @Override
-    public URI getLinkedContext() {
-        return linkedContext;
-    }
-
-    @Override
-    public void setLinkedContext(String linkedContext) {
-        setLinkedContext(URI.create(linkedContext));
-    }
-
-    protected void setLinkedContext(URI linkedContext) {
-        this.linkedContext = linkedContext;
     }
 
     public static final class Builder {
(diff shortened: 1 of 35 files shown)


src/main/java/no/unit/nva/model/contexttypes/Book.java
```diff
@@ -15,7 +17,7 @@
 import org.apache.commons.validator.routines.ISBNValidator;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
-public class Book implements BasicContext {
+public class Book implements BasicContext, LinkedContext {
```
axthosarouris, line 20: In an example of a BookAnthology and a Chapter, Would the Chapter have a linkedContext pointing to the BookAnthology ? If it is not this case but the opposite, we should have a collection of linkedContexts

axthosarouris (changes requested):

axthosarouris:
Requested change is only the docuementation.

axthosarouris:
If linkedContext is a URI why is the setter a String?

src/main/java/no/unit/nva/model/contexttypes/Book.java
```diff
@@ -164,10 +169,25 @@ public int hashCode() {
...
+    }
+
+    @Override
+    public void setLinkedContext(String linkedContext) {
+        setLinkedContext(URI.create(linkedContext));
+    }
+
+    protected void setLinkedContext(URI linkedContext) {
+        this.linkedContext = linkedContext;
+    }
```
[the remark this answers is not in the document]
svenngjeto: Getters and setters reorganized together as they should have been...

src/main/java/no/unit/nva/model/contexttypes/Book.java
svenngjeto: Added an sample with a Book in a series, but there will/must be upcoming changes in the model regardin handling of series

src/main/java/no/unit/nva/model/contexttypes/LinkedContext.java
```diff
@@ -4,7 +4,22 @@
 
 /**
  * LinkedContexts are contexts that are expressed as a URI, allowing an existing (external) Publication
- * to be used as a PublicationContext. For example, a BookAnthology containing a Chapter.
+ * to be used as a PublicationContext. For example, a BookAnthology containing a Chapter or a Book as part of a series.
+ *
```
axthosarouris, line 8: This does not answer the basic question which is if the container has links to the contained or the contained a link to the container. A sentence of the form "e.g. A Chapter will have a linkedContext to the BookAnthology that contains it" or "A Book Anthology will have a linkedContext to the Chapter". I have understood that it must be the former but one cannot be too clear.

axthosarouris (changes requested):

src/main/java/no/unit/nva/model/contexttypes/LinkedContext.java
svenngjeto: Tried to formulate some smart sentences to describe the funtionality

src/main/java/no/unit/nva/model/contexttypes/Book.java
brinxmat: The linkedContext is  always X isPartOf Y, not Y hasPart X

brinxmat: Moreover, if we want to represent these things in terms of the thing that has the hasPart relationship, I consider this to be an index task as there is no guarantee that that we know the parts (example, the editor of an anthology, where all the contributors are not at Norwegian institutions)

brinxmat (approved):

axthosarouris (approved):

PR #174, opened 2021-08-11
bugfix:tolerate npi level misspelling

axthosarouris:


src/main/java/no/unit/nva/model/instancetypes/journal/JournalArticle.java
```diff
@@ -12,7 +13,9 @@
     private String volume;
     private String issue;
     private String articleNumber;
-    private JournalArticleContentType content;
+
+    @JsonAlias("content") //for backwards compatibility.
```
axthosarouris, line 17: Renamed for compatibility with the other Publication instances. Should have been in another PR, but slipped into this one.

src/test/java/no/unit/nva/model/instancetypes/journal/JournalArticleTest.java
```diff
@@ -96,8 +97,8 @@ void journalArticleSerializationContainsJournalArticleContentType() throws JsonP
         JournalTestData journalTestData = new JournalTestData(RESEARCH_ARTICLE);
         JournalArticle journalArticle = generateJournalArticle(journalTestData);
         String json = objectMapper.writeValueAsString(journalArticle);
-        String expectedContentPhrase = "content\" : \"" + RESEARCH_ARTICLE.getValue() + "\"";
-        assertTrue(json.contains(expectedContentPhrase));
+        String expectedContentPhrase = "contentType\" : \"" + RESEARCH_ARTICLE.getValue() + "\"";
```
axthosarouris, line 100: fixed tests after rename "content" to "contentType"

src/main/java/no/unit/nva/model/Level.java
```diff
@@ -1,32 +1,61 @@
...
+import java.util.List;
+import java.util.Locale;
 import no.unit.nva.model.exceptions.InvalidNpiLevelException;
+import nva.commons.core.SingletonCollector;
 
 public enum Level {
-    LEVEL_2(2),
-    LEVEL_1(1),
-    LEVEL_0(0),
+    LEVEL_2(2, "level2", "level_2"),
```
brinxmat, line 14: Is this a list that will grow? If not, OK. Otherwise, keep "level" and 0-9, replace rest. match against "levelX"

brinxmat (approved):

PR #182, opened 2021-08-31
New property: subjects for URIs to controlled vocabularies

janbj:


janbj:
> Do these URIs point to NSD and will we need to derefence them?

@brinxmat sier nei.

brinxmat (approved):

PR #187, opened 2021-09-06
Np-3089 Make test and implement the removal of none-digits from isbn.

Eirik-Eide-Saur:


axthosarouris (changes requested):

axthosarouris (commented):
Approved but the name of the constant does not make sense. one basically says replace all digits with empty string

axthosarouris (approved):

PR #189, opened 2021-09-08
Make Book constructor public

brinxmat:
- Because of issues due to the public constructor requiring string and/or BookSeries-object, when creating new objects in a normal workflow, we end up with an unfortunate never-occurring situation (seriesTitle string will always be null


janbj (dismissed):

src/main/java/no/unit/nva/model/contexttypes/Book.java
```diff
@@ -2,6 +2,7 @@
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import no.unit.nva.model.contexttypes.utils.BookSeriesUtil;
```
axthosarouris, line 5: In line 103, in the copy() method, we can remove the `throws InvalidIsbnException`

src/main/java/no/unit/nva/model/contexttypes/utils/BookSeriesUtil.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private static boolean hasUnmatchedSeriesStringValues(BookSeries series, String unconfirmedSeriesTitle) {
+        return !series.isConfirmed()
+                && !((UnconfirmedSeries) series).getTitle().equals(unconfirmedSeriesTitle);
+    }
+
+    private static boolean hasSeriesStringAndSeriesObject(BookSeries series, String unconfirmedSeriesTitle) {
+        return nonNull(series) && nonNull(unconfirmedSeriesTitle);
+    }
+}
```
axthosarouris, line 44: Why not move all the methods as they are in in the interface BookSeries?

axthosarouris (changes requested):

axthosarouris (approved):

PR #193, opened 2021-09-15
add issn to UnconfirmedSeries

axthosarouris:
This PR adds ISSN field to Unconfirmed series, because Cristin has such entries. 
Changes:
* Add "issn" field and parameter to UnconfirmedSeries constructor.
* Make the constructor that accepts only a title a static method to avoid throwing an InvalidIssnException that cannot occur.
* Added a randomIssn() and randomInvalidIssn() method for creating random ISSNs

### DIFF
--- a/src/test/java/no/unit/nva/model/contexttypes/UnconfirmedSeriesTest.java
+++ b/src/test/java/no/unit/nva/model/contexttypes/UnconfirmedSeriesTest.java
@@ -7,46 +7,65 @@ import static nva.commons.core.JsonUtils.objectMapperWithEmpty;
 import static nva.commons.core.attempt.Try.attempt;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.core.Is.is;
 import static org.hamcrest.core.IsEqual.equalTo;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import com.fasterxml.jackson.databind.exc.ValueInstantiationException;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import no.unit.nva.model.exceptions.InvalidIssnException;
+import no.unit.nva.model.util.PublicationGenerator;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.function.Executable;
 
 class UnconfirmedSeriesTest {
 
-//    @Test
-//    public void constructorReturnsUnconfirmedSeriesWhenInputContainsValidIssn() throws InvalidIssnException {
-//        String randomIssn = randomIssn();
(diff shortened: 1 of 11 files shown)


src/test/java/no/unit/nva/model/util/PublicationGenerator.java
```diff
@@ -410,220 +421,260 @@ private static Publication generatePublicationWithReference(Reference reference)
...
+            .withOnlineIssn("1111-1119")
+            .withPrintIssn("2222-2227")
+            .build();
     }
 
     private static String randomWord() {
         return FAKER.lorem().word() + FAKER.lorem().word();
     }
+
+    private static String random7DigitNumber() {
```
brinxmat, line 641: Can we extract the ISSN generation code to a separate helper class?

src/test/java/no/unit/nva/model/contexttypes/UnconfirmedSeriesTest.java
```diff
@@ -0,0 +1,44 @@
...
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.core.Is.is;
+import static org.hamcrest.core.IsEqual.equalTo;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import no.unit.nva.model.exceptions.InvalidIssnException;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.function.Executable;
+
+class UnconfirmedSeriesTest {
```
brinxmat, line 18: Does this cover something not covered in more generic tests elsewhere?

axthosarouris: We do not cover the cases where the issns are invalid

axthosarouris: I checked and we do not have test coverage reports for the datamodel-java

src/test/java/no/unit/nva/model/util/PublicationGenerator.java
axthosarouris: Moved to a different class

src/test/java/no/unit/nva/model/contexttypes/UnconfirmedSeriesTest.java
axthosarouris: I have added a test for deserialzation of unconfirmed series.  I think trying to inject an invalid JSON UnconfirmedSeries object to a Publication JSON object, would require a lot of code and syncing between fields. In addition given the many objects that a Publication has and the many cases per object that must be tested, this could lead to an large number of tests and a need for complex data generation.

Example:
It is easier to test the de/serialization features of UnconfirmedSeries focusing only on this object and having tests such as 
"deserialization suceeds when input contains valid issn and valid online issn" 
"deserialization fails when input contains invalid issn"
than to test
"deserialization succeeds when a Publication refers to Book with Unconfirmed Series  that contains a valid issn and valid online issn"
"deserialization fails when a Publication refers to Book with Unconfirmed Series  that contains an valid issn."

brinxmat (approved):

PR #195, opened 2021-09-28
Add events

brinxmat:
- Adds support for conference lecture, conference poster and lecture
- We have a minor issue with "pages" as this now seems like it is drifting towards "extent", where extent may be a temporal extent (poster, lecture, although both may. have "pages" in terms of a single page (poster, but really dimensions are likely a better descriptor) and a slide deck), not just a number of pages. This will become even clearer in the next phase, where we introduce things that patently do not have "pages" (artistic results), so I suggest revisiting this in a new task before starting on artistic result


src/main/java/no/unit/nva/model/Agent.java
```diff
@@ -0,0 +1,11 @@
+package no.unit.nva.model;
+
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+@JsonSubTypes({
+    @JsonSubTypes.Type(name = "Organization", value = Organization.class)
+})
+public interface Agent {
```
axthosarouris, line 10: What is an Agent? Is it the institution through which the Resource has been published?

src/main/java/no/unit/nva/model/contexttypes/Event.java
```diff
@@ -0,0 +1,115 @@
...
+import no.unit.nva.model.Agent;
+import no.unit.nva.model.Publication;
+import no.unit.nva.model.contexttypes.event.Place;
+import no.unit.nva.model.pages.TemporalExtent;
+import nva.commons.core.JacocoGenerated;
+
+import java.util.Objects;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class Event implements PublicationContext {
```
axthosarouris, line 14: Is Event going to replace other PublicationContexts?

src/main/java/no/unit/nva/model/contexttypes/event/ReifiedExtent.java
```diff
@@ -0,0 +1,42 @@
+package no.unit.nva.model.contexttypes.event;
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+import java.util.Objects;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class ReifiedExtent {
```
axthosarouris, line 8: Is this an abstract representation of duration, e.g. 3 generations, 4 aeons, 2 lifetimes etc.?

src/main/java/no/unit/nva/model/pages/TemporalExtent.java
```diff
@@ -0,0 +1,53 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import no.unit.nva.model.contexttypes.event.ReifiedExtent;
+import nva.commons.core.JacocoGenerated;
+
+import java.time.LocalDateTime;
+import java.time.temporal.ChronoUnit;
+import java.util.Objects;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class TemporalExtent implements Pages {
```
axthosarouris, line 13: This class does not appear in the examples. How is LocalDateTime serialized?

axthosarouris:
As an implementation, the datamodel is unproblematic. However, I am not sure whether the review can or should extend to the structure of the conceptual model as presented through its implementation. 
Also, I think that we should put a date e.g.  31 January 2022, where we will require that the conceptual model is kept in sync with the implementation model, or we should find an alternative way to document the meaning of certain fields. E.g.  I am not sure what  "reifiedExtent" or "agent" are about  exactly. I think I understand their meaning, but I am not sure my understanding is the same with that of other people in the team.

axthosarouris (approved):

janbj (approved):

axthosarouris (approved):

PR #197, opened 2021-09-28
Add module for implemeting datamodel migration check

axthosarouris:
This is the fist step for setting up a migration check for the datamodel.
A description of the process can be found here: https://unit.atlassian.net/l/c/EWzdjE8Z

In this first step I just moved the code around to put the datamodel into a module


build.gradle
```diff
@@ -1,112 +1,6 @@
...
-    packageName(project.group)  // forces the package. Defaults to '${project.group}'
-    useJavaOutput() //forces the output to be Java
-    buildConfigField("String", "MODEL_VERSION", "\"${project.version}\"")
-}
-
-java {
-    withJavadocJar()
-    withSourcesJar()
+plugins{
+    id 'io.github.gradle-nexus.publish-plugin'
```
axthosarouris, line 2: these two plugins have to run in the root project

buildSrc/src/main/groovy/nvadatamodel.publish-maven.gradle
```diff
@@ -0,0 +1,76 @@
+plugins{
```
axthosarouris, line 1: Publishing in its own plugin, because we do not want to publish all modules

migration/build.gradle
```diff
@@ -0,0 +1,9 @@
+plugins {
```
axthosarouris, line 1: This will be the module where we will test whether the previous version and the current version are compatible

nva-datamodel-java/src/test/java/no/unit/nva/PublicationTest.java
```diff
@@ -41,7 +41,7 @@
 
     public static final String TIMESTAMP_REGEX = "[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\\.[0-9]+";
     public static final String SOME_TIMESTAMP = "2020-09-23T09:51:23.044996Z";
-    public static final String DOCUMENTATION_PATH_TEMPLATE = "documentation/%s.json";
+    public static final String DOCUMENTATION_PATH_TEMPLATE = "../documentation/%s.json";
```
axthosarouris, line 44: documentation folder  fix

brinxmat (approved):

PR #200, opened 2021-09-30
Add artistic content

brinxmat:
- Looks like a lot of code, but 2/3 is documentation JSON
- Adds classes for representation of artistic content
- Since they don't have pages, they have a null page object

### DIFF
--- a/nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/ArtisticDesign.java
+++ b/nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/ArtisticDesign.java
@@ -5,8 +5,11 @@ import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import no.unit.nva.model.contexttypes.venue.Venue;
 import nva.commons.core.JacocoGenerated;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+
+import static java.util.Objects.nonNull;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 public class ArtisticDesign implements PublicationContext {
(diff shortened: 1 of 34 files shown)


nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/ArtisticDesign.java
```diff
@@ -0,0 +1,44 @@
...
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class ArtisticDesign implements PublicationContext {
+    public static final String VENUES = "venues";
+
+    @JsonProperty(VENUES)
+    private final List<Venue> venues;
+
+    public ArtisticDesign(@JsonProperty(VENUES) List<Venue> venues) {
+        this.venues = venues;
```
axthosarouris, line 19: maybe we should have something like `Objects.nonNull(venues)? venues : Collections.emptyList()`?

nva-datamodel-java/src/test/java/no/unit/nva/model/util/PublicationGenerator.java
```diff
@@ -101,6 +115,50 @@ public static Publication generatePublication(String type) throws InvalidIssnExc
                                                                       InvalidUnconfirmedSeriesException {
         Reference reference;
         switch (type) {
+            case "ArtisticDesignClothingDesign":
```
axthosarouris, line 118: Maybe in the future we should consider something like this: https://stackoverflow.com/questions/347248/how-can-i-get-a-list-of-all-the-implementations-of-an-interface-programmatically ?

axthosarouris (approved):

janbj (approved):

PR #201, opened 2021-09-30
Refactor event instances to use nullpages

brinxmat:
- Use null pages for objects without pages

### DIFF
--- a/documentation/Lecture.json
+++ b/documentation/Lecture.json
@@ -96,17 +96,10 @@
       "doi" : "https://123123/213123.com",
       "publicationInstance" : {
         "type" : "Lecture",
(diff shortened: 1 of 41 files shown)


documentation/Lecture.json
```diff
@@ -96,17 +96,10 @@
...
-          "introduction" : {
-            "type" : "Range",
-            "begin" : "i",
-            "end" : "ix"
-          },
-          "pages" : "321",
-          "illustrated" : true
-        },
-        "peerReviewed" : false
+          "type" : "NullPages"
```
axthosarouris, line 101: Is it always the case that a Lecture does not have pages?

axthosarouris (approved):

PR #202, opened 2021-10-01
NP-3204 Fix isbn validator.

Eirik-Eide-Saur:
No idea why all the json files where included but they became modified one way or another, same goes for the schema.yaml file

### DIFF
--- a/nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/Book.java
+++ b/nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/Book.java
@@ -26,7 +26,7 @@ public class Book implements BasicContext {
     public static final String JSON_PROPERTY_SERIES_NUMBER = "seriesNumber";
     public static final String JSON_PROPERTY_PUBLISHER = "publisher";
     public static final String JSON_PROPERTY_ISBN_LIST = "isbnList";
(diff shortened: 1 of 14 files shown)


brinxmat (approved):

axthosarouris (changes requested):

nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/Book.java
```diff
@@ -85,7 +85,7 @@ public PublishingHouse getPublisher() {
             return Collections.emptyList();
         }
         List<String> validIsbns = isbnList.stream()
-            .map(isbn -> isbn.replaceAll(ANYTHING_BUT_DIGITS_REGEX, ""))
+            .map(isbn -> isbn.replaceAll(ANYTHING_BUT_DIGITS_AND_X_REGEX, ""))
```
brinxmat, line 88: This is entirely wrong as we lose validation for ISBNs that are input manually. The correct approach here is to catch the validation error and resubmit without the ISBN if-and-only-if you really want to throw away data willy-nilly. The issue should really be fixed in the data source

brinxmat (changes requested):

axthosarouris (approved):

PR #204, opened 2021-10-06
Adding id to publication dto

janbj:
... which is not really named PubicationDto but PublicationResponse


axthosarouris (commented):
We also need a test that verifies that the serialization of PublicationResponse and other classes that implement Id contain the field "id"

nva-datamodel-java/src/main/java/no/unit/nva/PublicationMapper.java
```diff
@@ -143,4 +149,12 @@ private static void mapInternal(WithInternal request, Publication publication) {
         publication.setDoi(request.getDoi());
         publication.setDoiRequest(request.getDoiRequest());
     }
+
+    public static URI toId(SortableIdentifier identifier) {
```
axthosarouris, line 153: delete unused method

axthosarouris (changes requested):

PR #209, opened 2021-10-08
new commons and organizing imports

axthosarouris:


nva-datamodel-java/src/main/java/no/unit/nva/DatamodelConfig.java
```diff
@@ -0,0 +1,9 @@
+package no.unit.nva;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import nva.commons.core.JsonUtils;
+
+public class DatamodelConfig {
+
+    public static final ObjectMapper dataModelObjectMapper = JsonUtils.dtoObjectMapper;
```
axthosarouris, line 8: The new one and only mapper used in the project.

nva-datamodel-java/src/test/java/no/unit/nva/model/DoiRequestMessageTest.java
```diff
@@ -40,11 +38,11 @@ public void doiRequestContainsDoiRequestMessages() {
     @Test
     public void doiRequestMessageCanBeSerialized() {
         DoiRequestMessage message = sampleDoiRequestMessage();
-        JsonNode json = JsonUtils.objectMapper.convertValue(message, JsonNode.class);
+        JsonNode json = dataModelObjectMapper.convertValue(message, JsonNode.class);
 
         List<String> fieldNames = extractFieldNames();
 
-        assertThat(message, doesNotHaveNullOrEmptyFields());
+        assertThat(message, doesNotHaveEmptyValues());
```
axthosarouris, line 45: replace depreacted method in favor for new method.

PR #211, opened 2021-10-12
return empty collections instead of null

axthosarouris:


nva-datamodel-java/src/main/java/no/unit/nva/api/PublicationResponse.java
```diff
@@ -155,49 +156,49 @@ public void setPublisher(Organization publisher) {
...
-    public void setDoi(URI doi) {
-        this.doi = doi;
+    public void setEntityDescription(EntityDescription entityDescription) {
+        this.entityDescription = entityDescription;
     }
 
     @Override
-    public DoiRequest getDoiRequest() {
-        return doiRequest;
+    public List<ResearchProject> getProjects() {
```
axthosarouris, line 190: return empty list.

nva-datamodel-java/src/main/java/no/unit/nva/api/PublicationResponse.java
```diff
@@ -155,49 +156,49 @@ public void setPublisher(Organization publisher) {
...
-        this.doiRequest = doiRequest;
-        this.doiRequested = Objects.nonNull(doiRequest);
+    public void setProjects(List<ResearchProject> projects) {
+        this.projects = projects;
     }
 
     @Override
     public List<URI> getSubjects() {
-        return subjects;
+        return Objects.nonNull(subjects) ? subjects : Collections.emptyList();
```
janbj, line 201: Could also consider using Optional with .else(Collections.emptyList()) but this is fine.

janbj (approved):

PR #216, opened 2021-10-14
Build a project that can can depend on any datamodel version

axthosarouris:
This is a step towards implementing the approach described here: https://unit.atlassian.net/l/c/xxCGJgBu
In this PR we have done the following:
1.We have created a nested project that is used to produce datamodel serializations of any datamodel version
2. We have created a functional test that builds the sample project for the current (published or unpublished) version.

The current produced Publications are incomplete. The focus is to be able to create a sample Publication serialization of a specified version.

### DIFF
--- a/sample-generation/gradle/libs.versions.toml
+++ b/sample-generation/gradle/libs.versions.toml
@@ -1,3 +1,5 @@
+## This file is replaced during the Functional tests and its only purpose
+## is to make this project build independently.
 [versions]
 junitVersion = { strictly = '5.6.2' }
 jacksonVersion = { strictly = '2.13.0' }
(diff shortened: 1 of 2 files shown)


sample-generation/src/main/java/no/unit/nva/sampleproject/PublicationGenerator.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.List;
+import java.util.Map;
+import no.unit.nva.identifiers.SortableIdentifier;
+import no.unit.nva.model.Organization;
+import no.unit.nva.model.Publication;
+
+public class PublicationGenerator {
+
+    public static Publication createPublication() {
+        return new Publication.Builder()
```
axthosarouris, line 14: A small class for creating a random publication. This will be extended to create a complete list of all possible Resource types with all possible Publication Instances and Contexts

PR #234, opened 2021-12-06
create resource owner object

axthosarouris:


nva-datamodel-java/src/test/java/no/unit/nva/model/PublicationTest.java
```diff
@@ -92,9 +92,8 @@ public void equalsReturnsTrueWhenTwoPublicationInstancesHaveEquivalentFields() {
         assertThat(copy, doesNotHaveEmptyValuesIgnoringFields(Set.of(DOI_REQUEST_FIELD)));
 
         Diff diff = JAVERS.compare(samplePublication, copy);
-        assertThat(diff.hasChanges(), is(false));
```
axthosarouris, line 95: diff compares fields directly and it fails when we copy data because we copy through getters and getOwner is overriden by getReourceOwner()

janbj (dismissed):

janbj (approved):

PR #235, opened 2021-12-06
refactor publication constructor

axthosarouris:


buildSrc/src/main/groovy/nvadatamodel.java-conventions.gradle
```diff
@@ -54,7 +54,7 @@ pmd {
     ignoreFailures = false
     pmdMain {
         excludes = [
-                '**/BuildConfig.*'
+                '**/ResourcesBuildConfig.*'
```
axthosarouris, line 57: renamed because it conflicts with file-model's BuldConfig

brinxmat (approved):

PR #242, opened 2021-12-10
Update get doi requested

axthosarouris:


nva-datamodel-java/src/main/java/no/unit/nva/api/PublicationResponse.java
```diff
@@ -246,8 +247,8 @@ public void setSubjects(List<URI> subjects) {
         this.subjects = subjects;
     }
 
-    public Boolean getDoiRequested() {
-        return doiRequested;
+    public boolean isDoiRequested() {
```
axthosarouris, line 250: being null creates problems. should be either true or false

janbj (approved):

PR #247, opened 2022-01-19
Move venue to ArtisticDesign

brinxmat:
Looks like a lot of changes, but the majority is the documentation, which is unfortunate. I guess I will provide a fix for this in another PR.

The main change here is that Venue is moved from Artistic -> Artistic design.

The test is updated to ignore the deprecated owner field in Publication.

### DIFF
--- a/buildSrc/src/main/groovy/nvadatamodel.java-conventions.gradle
+++ b/buildSrc/src/main/groovy/nvadatamodel.java-conventions.gradle
@@ -9,7 +9,7 @@ plugins {
 }
 
 group 'com.github.bibsysdev'
-version '0.14.23'
+version '0.14.24'
 
 
 repositories {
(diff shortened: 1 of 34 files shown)


buildSrc/src/main/groovy/nvadatamodel.java-conventions.gradle
```diff
@@ -9,7 +9,7 @@ plugins {
 }
 
 group 'com.github.bibsysdev'
-version '0.14.23'
+version '0.14.24'
```
axthosarouris, line 12: Probably this is 0.15.0, since the new version is not compatible with the old one

axthosarouris (changes requested):
see comment about model version

janbj (dismissed):
Can we delete the generated json-files (or git ignore them), and just commit the schema.yaml on updates?

axthosarouris (approved):

PR #251, opened 2022-01-31
Add award

brinxmat:
- Adds the Award concept
- Covers prizes, public recognition and other merits


nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/PublicationInstanceBuilder.java
```diff
@@ -371,7 +372,12 @@ private static String randomVolume() {
     }
 
     private static List<ArchitectureOutput> randomArchitectureOutputs() {
-        return List.of(randomCompetition(), randomMentionInPublication());
+        return List.of(randomCompetition(), randomMentionInPublication(), randomAward());
+    }
+
+    private static ArchitectureOutput randomAward() {
+        return new Award(randomString(), randomString(), new Instant(LocalDateTime.now()), randomInteger(),
```
axthosarouris, line 379: we have randomInstant() and randomInstant(Instant after), it would be better to use that, because then we have a better control over the timestamps. E.g., in case where we would expect a certain date to be after another date, (e.g. modifiedDate should be after createdDate)

axthosarouris (changes requested):

axthosarouris (dismissed):

axthosarouris (dismissed):

axthosarouris (approved):

PR #253, opened 2022-01-31
Add sequence interface

brinxmat:
- Extraction of initial interface for common sequence concept
- A discussion is required about what is actually common across these classes


nva-datamodel-java/src/main/java/no/unit/nva/model/instancetypes/artistic/realization/Exhibition.java
```diff
@@ -23,17 +24,21 @@
...
     private final String otherInformation;
+    @JsonProperty(SEQUENCE)
+    private final int sequence;
 
     public Exhibition(@JsonProperty(NAME) String name,
                       @JsonProperty(PLACE) String place,
                       @JsonProperty(ORGANIZER) String organizer,
                       @JsonProperty(TIME) Period time,
-                      @JsonProperty(OTHER_INFORMATION) String otherInformation) {
+                      @JsonProperty(OTHER_INFORMATION) String otherInformation,
```
axthosarouris, line 34: I am not sure whether it is applicable here, but I will put a reminder here of our discussion that we do not want Strings but typed objects.

axthosarouris (approved):

PR #259, opened 2022-02-05
Add performing arts

brinxmat:
- Activates test for moving picture (error from previous PR)
- Adds performing arts

### DIFF
--- a/documentation/schema.yaml
+++ b/documentation/schema.yaml
@@ -128,13 +128,7 @@ referencedSchemas:
     - type: object
       properties:
         subtype:
-          type: string
-          enum:
-          - BUILDING
-          - PLANNING_PROPOSAL
-          - LANDSCAPE_ARCHITECTURE
-          - INTERIOR
-          - OTHER
+          $ref: '#/components/schemas/ArchitectureSubtype'
         description:
           type: string
         architectureOutput:
@@ -157,6 +151,20 @@ referencedSchemas:
         type: string
     discriminator:
       propertyName: type
+  ArchitectureSubtype:
+    type: object
+    properties:
+      type:
+        type: string
+        enum:
+        - BUILDING
+        - PLANNING_PROPOSAL
+        - LANDSCAPE_ARCHITECTURE
+        - INTERIOR
+        - OTHER
+      description:
+        type: string
+        writeOnly: true
   Artistic:
     required:
     - type
@@ -1357,10 +1365,10 @@ referencedSchemas:
     - type
     type: object
     properties:
-      peerReviewed:
-        type: boolean
       pages:
         $ref: '#/components/schemas/Pages'
+      peerReviewed:
+        type: boolean
       type:
         type: string
     discriminator:
--- a/documentation/schema.yaml
+++ b/documentation/schema.yaml
@@ -320,6 +320,22 @@ referencedSchemas:
         type: string
     discriminator:
       propertyName: type
+  Broadcast:
+    required:
+    - type
+    type: object
+    discriminator:
+      propertyName: type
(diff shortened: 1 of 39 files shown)


nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/PublicationContextBuilder.java
```diff
@@ -42,7 +42,9 @@ public static PublicationContext randomPublicationContext(Class<?> publicationIn
         switch (className) {
             case "Architecture":
             case "ArtisticDesign":
-                return randomArtisticDesign();
+            case "MovingPicture":
```
brinxmat, line 45: MovingPicture was left out of the previous PR

nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/PublicationInstanceBuilder.java
```diff
@@ -84,7 +89,9 @@
         var typeName = randomType.getSimpleName();
 
         switch (typeName) {
-            case "MotionPicture":
+            case "PerformingArts":
+                return generatePerformingArts();
+            case "MovingPicture":
```
brinxmat, line 94: MovingPicture was left out of the previous PR

documentation/schema.yaml
```diff
@@ -320,6 +320,22 @@ referencedSchemas:
         type: string
     discriminator:
       propertyName: type
+  Broadcast:
```
brinxmat, line 323: MovingPicture was not added to the previous PR because the test code was not activated

nva-datamodel-java/src/main/java/no/unit/nva/model/instancetypes/artistic/performingarts/PerformingArtsSubtype.java
```diff
@@ -0,0 +1,59 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import nva.commons.core.JacocoGenerated;
+
+import java.util.Objects;
+
+public class PerformingArtsSubtype {
+    public static final String TYPE = "type";
+    public static final String DESCRIPTION = "description";
+    @JsonProperty(TYPE)
```
axthosarouris, line 12: To my understanding, this is a way to overcome some feature that Jackson does not cover. Can we indicate this somehow? We could for example create an interface with a name "VariablyTyped" that would expect `getType` to be implemented and would also contain a Javadoc comment explaining the role of this interface. The reaason I am saying this, is to avoid the temptation of changing the field name, or if we ever use another serialization library, to maybe address the problem differently.

nva-datamodel-java/src/main/java/no/unit/nva/model/instancetypes/artistic/performingarts/PerformingArtsSubtypeEnum.java
```diff
@@ -0,0 +1,20 @@
+package no.unit.nva.model.instancetypes.artistic.performingarts;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+public enum PerformingArtsSubtypeEnum {
+    THEATRICAL_PRODUCTION("TheatricalProduction"),
```
axthosarouris, line 6: Thinking about AWS SDK2, and its problems with additional labels, is there a reason to have additional labels? Is it to be consistent with the types produced by Jackson (i.e., class names)?

axthosarouris (dismissed):

PR #261, opened 2022-02-09
Add literaryarts

brinxmat:
- Adds support for literary arts


nva-datamodel-java/src/main/java/no/unit/nva/model/instancetypes/artistic/film/MovingPictureSubtypeEnum.java
```diff
@@ -1,8 +1,9 @@
 package no.unit.nva.model.instancetypes.artistic.film;
 
 import com.fasterxml.jackson.annotation.JsonValue;
+import no.unit.nva.model.instancetypes.artistic.WithType;
 
-public enum MovingPictureSubtypeEnum {
+public enum MovingPictureSubtypeEnum implements WithType {
```
axthosarouris, line 6: This enum returns the "type" field as its value but it lacks the JsonCreator for parsing the type.

axthosarouris, line 6: I do not think that the enum implements an interface WithType but an interface IsType.

janbj (dismissed):

nva-datamodel-java/src/main/java/no/unit/nva/model/instancetypes/artistic/literaryarts/LiteraryArtsSubtypeEnum.java
```diff
@@ -0,0 +1,29 @@
...
+    SHORT_STORY("ShortStory"),
+    ESSAY("Essay"),
+    TRANSLATION("Translation"),
+    RETELLING("Retelling"),
+    PLAY("Play"),
+    OTHER("Other");
+
+
+    @JsonValue
+    private final String type;
```
axthosarouris, line 19: Same problem with de/serializing

nva-datamodel-java/src/main/java/no/unit/nva/model/instancetypes/artistic/film/MovingPictureSubtypeEnum.java
```diff
@@ -17,6 +18,7 @@
         this.type = type;
     }
 
+    @Override
```
axthosarouris, line 21: I am not sure whether you want the subtype to be an object or a String. Now it is like this:
```
"publicationInstance" : {
        "type" : "MovingPicture",
        "subtype" : {
          "type" : "InteractiveFilm"
        },
```

If you want it to be a String i.e.,
``` 
    "publicationInstance" : {
        "type" : "MovingPicture",
        "subtype" : "InteractiveFilm"
```

then you have to add a structure similar to the following:
```
@JsonCreator
    public static MovingPictureSubtype fromString(String type){
        
    }
    
    @JsonValue
    public String toString(){
        this.getType();
    }
``` 
and remove the annotation from the field

nva-datamodel-java/src/main/java/no/unit/nva/model/instancetypes/artistic/literaryarts/LiteraryArtsSubtype.java
```diff
@@ -0,0 +1,59 @@
+package no.unit.nva.model.instancetypes.artistic.literaryarts;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import nva.commons.core.JacocoGenerated;
+
+import java.util.Objects;
+
+public class LiteraryArtsSubtype {
```
axthosarouris, line 9: Similar problem:

```
ublicationInstance" : {
        "type" : "LiteraryArts",
        "subtype" : {
          "type" : "Poetry"
        },
```

PR #262, opened 2022-02-11
Deprecated owner field

janbj:


brinxmat (approved):

nva-datamodel-java/src/main/java/no/unit/nva/model/Publication.java
```diff
@@ -116,13 +115,15 @@ public void setModifiedDate(Instant modifiedDate) {
     }
 
     @Override
+    @Deprecated
     public String getOwner() {
-        return Optional.ofNullable(resourceOwner).map(ResourceOwner::getOwner).orElse(owner);
+        return null;
```
axthosarouris, line 120: I would suggest that we either delete the owner field or we maintain its value and annotate it as @Deprecated. Making it null, will cause a lot of runtime errors instead of compilation errors.

axthosarouris: Making it null and not deleting the getter and setter breaks many things.

axthosarouris (changes requested):

PR #267, opened 2022-05-02
Added new Data model ConferenceAbstract

RanjithTom:


nva-datamodel-java/src/main/java/no/unit/nva/model/instancetypes/journal/ConferenceAbstract.java
```diff
@@ -0,0 +1,65 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import no.unit.nva.model.pages.Range;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class ConferenceAbstract extends JournalNonPeerReviewedContent {
+
+    /**
+     * Creates an object that matches the Journal pattern, but throws an exception if peerReviewed is set to true.
```
axthosarouris, line 12: Not informative comment. It does not describe what is a ConferenceAbstract.

axthosarouris (changes requested):
see comment

brinxmat (dismissed):
Fix the code comment, otherwise OK — I still don't understand what this is actually supposed to represent, but I guess I'm not alone in this.

axthosarouris (approved):

PR #274, opened 2022-05-16
Bump version

axthosarouris:


buildSrc/src/main/groovy/nvadatamodel.java-conventions.gradle
```diff
@@ -9,7 +9,7 @@ plugins {
 }
 
 group 'com.github.bibsysdev'
-version '0.16.8'
+version '0.16.9'
```
axthosarouris, line 12: This the only change I made. The rest are json files.

brinxmat (approved):

PR #287, opened 2022-06-16
NP-9172: Add person cristin id to Identity

torarnet:


nva-datamodel-java/src/main/java/no/unit/nva/model/Identity.java
```diff
@@ -65,27 +68,36 @@ public void setArpId(String arpId) {
         this.arpId = arpId;
     }
 
+    public URI getPersonCristinId() {
+        return personCristinId;
```
brinxmat, line 72: id == cristinId, in fact, it isn't important where the URI comes from…but canonically, it should be an rdf:Resource of type nva:Person

brinxmat (changes requested):

nva-datamodel-java/src/main/java/no/unit/nva/model/Identity.java
torarnet: So I should use `private URI id;` and set that to Cristin proxy id? Then deleting arpId and not adding anymore fields?

PR #289, opened 2022-07-13
Np 9300 add performing arts roles

brinxmat:


elkebab (dismissed):

brinxmat:
Changes carried over in subsequent PR

PR #293, opened 2022-09-01
Add part of series boolean

brinxmat:
Adds a simple boolean property to concert types to indicate that they are part of a concert series/tour.

### DIFF
--- a/nva-datamodel-java/src/main/java/no/unit/nva/model/instancetypes/artistic/music/Concert.java
+++ b/nva-datamodel-java/src/main/java/no/unit/nva/model/instancetypes/artistic/music/Concert.java
@@ -11,34 +11,39 @@ import nva.commons.core.JacocoGenerated;
 
 public class Concert implements MusicPerformanceManifestation {
 
-    public static final String PLACE = "place";
-    public static final String TIME = "time";
-    public static final String EXTENT = "extent";
-    public static final String DESCRIPTION = "description";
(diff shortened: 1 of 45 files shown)


nva-datamodel-java/src/main/java/no/unit/nva/model/instancetypes/artistic/music/Concert.java
```diff
@@ -11,34 +11,39 @@
...
+    @JsonProperty(EXTENT_FIELD)
     private final String extent;
-    @JsonProperty(DESCRIPTION)
+    @JsonProperty(DESCRIPTION_FIELD)
     private final String description;
-    @JsonProperty(CONCERT_PROGRAMME)
+    @JsonProperty(CONCERT_PROGRAMME_FIELD)
     private final List<MusicalWorkPerformance> concertProgramme;
+    @JsonProperty(PART_OF_SERIES_FIELD)
+    private final boolean partOfSeries;
```
torarnet, line 32: primitive? false if not set?

torarnet (approved):
Approved with comment

nva-datamodel-java/src/main/java/no/unit/nva/model/instancetypes/artistic/music/Concert.java
brinxmat: I'm assuming it is false if not set

PR #300, opened 2022-09-28
Add cloned files from FileSet as associatedArtifacts

brinxmat:
The first step in migrating from filesets to associated artifacts

### DIFF
--- a/nva-datamodel-java/src/main/java/no/unit/nva/model/Publication.java
+++ b/nva-datamodel-java/src/main/java/no/unit/nva/model/Publication.java
@@ -1,6 +1,7 @@
 package no.unit.nva.model;
 
 import static java.util.Objects.hash;
+import static java.util.Objects.nonNull;
 import static nva.commons.core.attempt.Try.attempt;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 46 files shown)


nva-datamodel-java/src/main/java/no/unit/nva/model/Publication.java
```diff
@@ -317,7 +319,11 @@ private void verifyStatusTransition(PublicationStatus nextStatus)
             throw new InvalidPublicationStatusTransitionException(currentStatus, nextStatus);
         }
     }
-    
+
+    public List<File> getAssociatedArtifacts() {
+        return new ArrayList<>(getFileSet().getFiles());
```
axthosarouris, line 324: if getFileSet returns null you will get a null pointer exception

axthosarouris (changes requested):

nva-datamodel-java/src/main/java/no/unit/nva/model/Publication.java
brinxmat: Fixed, good catch

axthosarouris (approved):

PR #312, opened 2022-10-21
Fix serialization and add associated artifacts null object

brinxmat:


nva-datamodel-java/src/main/java/no/unit/nva/model/associatedartifacts/NullAssociatedArtifact.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.annotation.JsonTypeInfo.Id;
+
+/**
+ * This class represents a state for publications that do not and cannot have
+ * associated artifacts.
+ */
+@JsonTypeInfo(use = Id.NAME, property = "type")
+public class NullAssociatedArtifact implements AssociatedArtifact {
```
axthosarouris, line 11: add a type explicitly with `@JsonTypeName`?

nva-datamodel-java/src/main/java/no/unit/nva/model/instancetypes/artistic/film/realization/MovingPictureOutput.java
```diff
@@ -6,9 +6,9 @@
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 @JsonSubTypes({
-        @JsonSubTypes.Type(name = "Broadcast", value = Broadcast.class),
-        @JsonSubTypes.Type(name = "CinematicRelease", value = CinematicRelease.class),
-        @JsonSubTypes.Type(name = "OtherRelease", value = OtherRelease.class)
+    @JsonSubTypes.Type(name = "Broadcast", value = Broadcast.class),
```
axthosarouris, line 9: General comment: Better if the type name is contained the class and then you can have 
```
@JsonSubTypes.Type(name = Broadcast.TYPE, value = Broadclas.class)
```

axthosarouris (approved):

PR #320, opened 2022-10-30
Apply model verification for publication context

brinxmat:
- Adds a method to get the JSON-LD context object to the Publication object
- Verifies that the context object is to some extent valid


nva-datamodel-java/src/test/java/no/unit/nva/PublicationContextTest.java
```diff
@@ -0,0 +1,199 @@
...
+    private Set<String> generateSetOfPropertiesThatHaveCollectionTypeForEveryNvaType() {
+        return publicationInstanceProvider()
+                .map(PublicationGenerator::randomPublication)
+                .map(PublicationContextTest::convertToJsonNode)
+                .map(PublicationContextTest::initiateArrayFieldNameExtraction)
+                .flatMap(Collection::stream)
+                .collect(Collectors.toSet());
+    }
+
+    private static Set<String> initiateArrayFieldNameExtraction(JsonNode jsonNode) {
```
brinxmat, line 114: Tail function that needs init

PR #321, opened 2022-11-01
Add ontology to data model

brinxmat:
Adds the ontology to the model, we will need to address the issues with the ontology later, for now it is enough to have the basic mappings.

There is an obvious need for simplification here.

SEE NOTE BELOW: If you want to have fun, to see the work ahead, take a look at https://service.tib.eu/webvowl/#iri=https://raw.githubusercontent.com/BIBSYSDEV/nva-datamodel-java/add-ontology-to-data-model/nva-datamodel-java/src/main/resources/publication-ontology.ttl

NOTE: Use Filter > Degree of collapsing > 0


nva-datamodel-java/src/main/resources/publication-ontology.ttl
```diff
@@ -0,0 +1,1744 @@
+@prefix nva: <https://nva.sikt.no/ontology/publication#> .
```
torarnet, line 1: Did you write all this manually?

nva-datamodel-java/src/test/java/no/unit/nva/OntologyTest.java
```diff
@@ -0,0 +1,159 @@
...
+    @Test
+    void shouldContainEveryVisibleClassOfModel() {
+        var ontologyClasses = extractClassesFromOntology();
+        var modelClasses = new ArrayList<>(getModelClasses()).toArray(String[]::new);
+        assertThat(ontologyClasses, hasItems(modelClasses));
+    }
+
+    @Test
+    void shouldContainEveryVisiblePropertyOfModel() {
+        var ontologyClasses = extractPropertiesFromOntology();
```
torarnet, line 76: Names instead ontologyProperties and modelProperties? Not important though

torarnet (dismissed):

ketilaa (dismissed):

nva-datamodel-java/src/test/java/no/unit/nva/OntologyTest.java
brinxmat: Good catch

nva-datamodel-java/src/main/resources/publication-ontology.ttl
brinxmat: Nope, I generated the majority of it — but had to do a lot of small fixes manually

brinxmat: In the future, I suspect we will have some annotations that aid generation.

torarnet (approved):

PR #328, opened 2022-11-20
Add status for published metadata

brinxmat:
- Adding, reluctantly, a status where only metadata is published
- This status is used as an interim status when a publishing process is only partially completed, and full PUBLISHED status is a logical transition from this status

### DIFF
--- a/nva-datamodel-java/src/main/java/no/unit/nva/model/PublicationStatus.java
+++ b/nva-datamodel-java/src/main/java/no/unit/nva/model/PublicationStatus.java
@@ -9,6 +9,7 @@ public enum PublicationStatus {
 
     NEW("NEW"),
     DRAFT("DRAFT"),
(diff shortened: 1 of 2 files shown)


nva-datamodel-java/src/main/java/no/unit/nva/model/PublicationStatus.java
```diff
@@ -9,6 +9,7 @@ public enum PublicationStatus {
 
     NEW("NEW"),
     DRAFT("DRAFT"),
+    PUBLISHED_METADATA("PUBLISHED_METADATA"),
```
elkebab, line 12: Vil ikke dette i praksis fjerne behovet for `PUBLISHED`?
Antar klienten fortsatt skal forholde seg til `PublishedFile`/`UnpublishedFile` for filer, så `status` vil uansett representere metadata slik jeg forstår det

ketilaa (approved):

PR #330, opened 2022-12-09
added fundings to publication data model

ketilaa:
Made assumption that Cristin Proxy will produce a URI id for funding sources from Cristin. Did not take into account confirmed vs unconfirmed funding (projectId, title, amount), e.g. confirmed when taken from Forskningsrådet, and unconfirmed when taken as input from frontend.

### DIFF
--- a/documentation/schema.yaml
+++ b/documentation/schema.yaml
@@ -10,6 +10,7 @@ schema:
       enum:
       - NEW
       - DRAFT
+      - PUBLISHED_METADATA
       - PUBLISHED
       - DRAFT_FOR_DELETION
     resourceOwner:
@@ -43,6 +44,10 @@ schema:
       type: array
       items:
         $ref: '#/components/schemas/ResearchProject'
+    fundings:
+      type: array
+      items:
+        $ref: '#/components/schemas/Funding'
     additionalIdentifiers:
       uniqueItems: true
       type: array
@@ -56,9 +61,9 @@ schema:
     associatedArtifacts:
       type: array
       properties:
-        empty:
-          type: boolean
         publishable:
+          type: boolean
+        empty:
           type: boolean
       items:
         $ref: '#/components/schemas/AssociatedArtifact'
@@ -246,9 +251,9 @@ referencedSchemas:
   AssociatedArtifactList:
     type: array
     properties:
-      empty:
-        type: boolean
       publishable:
+        type: boolean
+      empty:
         type: boolean
     items:
       $ref: '#/components/schemas/AssociatedArtifact'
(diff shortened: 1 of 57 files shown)


documentation/schema.yaml
```diff
@@ -968,6 +973,27 @@ referencedSchemas:
...
           type: string
+  Funding:
+    required:
+    - type
+    type: object
+    properties:
+      source:
+        type: string
+        format: uri
+      title:
```
torarnet, line 984: We can have titles in more than one language can we not?

ketilaa: The UI design will not support it, but we can add it and assume Norwegian?

ketilaa: Or use mainTitle and provided ways of setting alternative titles in a map like for publication?

torarnet: We can still have a language map in Cristin proxy, and then the user of the proxy can choose which language themselves?

nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/RandomCurrencyUtil.java
```diff
@@ -0,0 +1,19 @@
+package no.unit.nva.model.testing;
+
+import java.util.Currency;
```
torarnet, line 3: Is this built-in in java?

torarnet (dismissed):

documentation/schema.yaml
ketilaa: Note that for standalone fundings for registrations that I am working on, we don't store title for funding sources. Only for the project: 
Screenshot 2022-12-09 at 14 36 50

ketilaa: Hmm, do we have to write these to Cristin so they can be reused for different publications?

nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/RandomCurrencyUtil.java
ketilaa: Yes, just to see familiar currency values in tests. Not considered how we want to validate user input in the API yet.

documentation/schema.yaml
```diff
@@ -56,10 +61,10 @@ schema:
     associatedArtifacts:
       type: array
       properties:
-        empty:
-          type: boolean
         publishable:
           type: boolean
+        empty:
+          type: boolean
```
brinxmat, line 67: Hvor kommer dette fra?

brinxmat (approved):

ketilaa:
Talked to PO, and did some adjustments accordingly.

documentation/schema.yaml
```diff
@@ -1594,6 +1621,19 @@ referencedSchemas:
           type: string
         type:
           type: string
+  MonetaryAmount:
+    type: object
+    properties:
+      currency:
+        type: string
+        enum:
+        - NOK
```
torarnet, line 1630: Tror cristin bare støtter NOK for øyeblikket, men det kan fort endre seg

documentation/schema.yaml
```diff
@@ -968,6 +973,28 @@ referencedSchemas:
...
+        type: string
+        format: uri
+      identifier:
+        type: string
+      mainTitle:
+        type: string
+      alternativeTitles:
+        type: object
+        additionalProperties:
+          type: string
```
torarnet, line 991: Does this match a map with key value strings?

torarnet (approved):

brinxmat (approved):

PR #331, opened 2022-12-28
adjustments to model for fundings

ketilaa:


nva-datamodel-java/src/main/java/no/unit/nva/model/Funding.java
```diff
@@ -10,10 +11,13 @@
...
+    private URI id;
     private String identifier;
-    private String mainTitle;
-    private Map<String, String> alternativeTitles;
-    private MonetaryAmount amount;
+    private String name;
+    private Map<String, String> alternativeName;
+    private MonetaryAmount fundingAmount;
+    private Instant activeFrom;
+    private Instant activeTo;
```
brinxmat, line 20: Quick question: The funding itself has an @id, which entails that every property relates to the object dereferenced by the @id. Is this the case, such that the @id always and only has the same values for e.g. funding amount, active*?

nva-datamodel-java/src/test/java/no/unit/nva/OntologyTest.java
```diff
@@ -64,18 +66,27 @@ void shouldContainDistinctDescriptions() {
...
+    @ParameterizedTest
+    @MethodSource("provideModelPropertiesArguments")
+    void shouldContainEveryVisiblePropertyOfModel(String modelProperty) {
         var ontologyProperties = extractPropertiesFromOntology();
-        var modelProperties = new ArrayList<>(getModelProperties()).toArray(String[]::new);
-        assertThat(ontologyProperties, hasItems(modelProperties));
+        assertThat(ontologyProperties, hasItem(modelProperty));
+    }
+
+    public static Stream<Arguments> provideModelPropertiesArguments() {
```
brinxmat, line 84: Call this modelPropertyProvider and keep it at the top of the class — this convention should probably be enforced by some rule, but isn't, alas.

nva-datamodel-java/src/test/java/no/unit/nva/OntologyTest.java
```diff
@@ -64,18 +66,27 @@ void shouldContainDistinctDescriptions() {
         assertThat(duplicatesMessage, distinctValues, is(equalTo(ontologyValues)));
     }
 
-    @Test
-    void shouldContainEveryVisibleClassOfModel() {
+    @ParameterizedTest
+    @MethodSource("provideModelClassesArguments")
```
brinxmat, line 70: Yep, this is a more constructive way of doing this :) call the method modelClassProvider.

nva-datamodel-java/src/test/java/no/unit/nva/OntologyTest.java
```diff
@@ -64,18 +66,27 @@ void shouldContainDistinctDescriptions() {
...
-        var modelProperties = new ArrayList<>(getModelProperties()).toArray(String[]::new);
-        assertThat(ontologyProperties, hasItems(modelProperties));
+        assertThat(ontologyProperties, hasItem(modelProperty));
+    }
+
+    public static Stream<Arguments> provideModelPropertiesArguments() {
+        return Stream.of(Arguments.of(getModelProperties().toArray()));
+    }
+
+    public static Stream<Arguments> provideModelClassesArguments() {
```
brinxmat, line 88: Also this should conventionally be at the top of the class.

brinxmat (changes requested):

nva-datamodel-java/src/main/java/no/unit/nva/model/Funding.java
ketilaa: Not really. If source is NFR, activeFrom and activeTo can be resolved by the `id`. fundingAmount is something that can be set per use of the funding, e.g. to indicate that a funding is split on different projects. Other verified funding sources we integrate to in the future might not even have `activeFrom` and `activeTo`.

ketilaa: I think that from publication api perspective, a funding should have either an `id` or a `name`. The rest of the fields should be optional. `name` is used when the data is entered manually in the form.

ketilaa: Another aspect worth mentioning is that I have added source URI to verified fundings, so the `source` is also given if an `id` is provided.

nva-datamodel-java/src/test/java/no/unit/nva/OntologyTest.java
ketilaa: fixed

nva-datamodel-java/src/test/java/no/unit/nva/OntologyTest.java
ketilaa: fixed

nva-datamodel-java/src/test/java/no/unit/nva/OntologyTest.java
ketilaa: And yes, it is easier to see what class/property is the problem this way.

nva-datamodel-java/src/test/java/no/unit/nva/OntologyTest.java
ketilaa: fixed

nva-datamodel-java/src/main/java/no/unit/nva/model/Funding.java
ketilaa: @brinxmat Do we want to change something based on this insight?

PR #334, opened 2023-01-11
NP-28714-possible-to-delete-published-publications, added enum

anetteOlli:
Høres denne enumen ok ut?

Se også forøvrig slack https://sikt-no.slack.com/archives/CN005AQCE/p1673430958877679

### DIFF
--- a/nva-datamodel-java/src/main/java/no/unit/nva/model/PublicationStatus.java
+++ b/nva-datamodel-java/src/main/java/no/unit/nva/model/PublicationStatus.java
@@ -11,6 +11,7 @@ public enum PublicationStatus {
     DRAFT("DRAFT"),
     PUBLISHED_METADATA("PUBLISHED_METADATA"),
     PUBLISHED("PUBLISHED"),
+    DELETED("DELETED"),
     DRAFT_FOR_DELETION("DRAFT_FOR_DELETION");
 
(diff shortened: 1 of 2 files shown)


nva-datamodel-java/src/main/java/no/unit/nva/model/PublicationStatus.java
```diff
@@ -11,6 +11,7 @@ public enum PublicationStatus {
     DRAFT("DRAFT"),
     PUBLISHED_METADATA("PUBLISHED_METADATA"),
     PUBLISHED("PUBLISHED"),
+    DELETED("DELETED"),
```
brinxmat, line 14: So it's clear, "Deleted" implies that 
- there was a Published resource that is now marked as deleted and should not be understood as published
- it does not appear in any search index
- its landing page is a "tombstone", stating that it has ceased to be

brinxmat (approved):

nva-datamodel-java/src/main/java/no/unit/nva/model/PublicationStatus.java
anetteOlli: Also see thread in https://sikt-no.slack.com/archives/CN005AQCE/p1673430958877679.

truhacevkir (approved):

PR #342, opened 2023-01-19
Allow only identifiable Organizations

brinxmat:
- Forces construction of identifiable organizations (ones with URIs)
- Removes underspecified organizations


nva-datamodel-java/src/test/java/no/unit/nva/MigrationTestingOrganizationTest.java
```diff
@@ -0,0 +1,29 @@
+package no.unit.nva;
```
ketilaa, line 1: How will returning `null` behave when we are parsing a list of `Organization`s?

ketilaa (commented):
What code is it the Brage import uses to add affiliations for contributors? Or do we just let them provide faulty data and fixes it on the way to the database?

ketilaa (approved):

PR #348, opened 2023-02-02
Migrate subtypes of journal article to top-level types

brinxmat:
- Migrates existing JournalArticle type into new classes representing the former content types of journal article
- Given a JournalArticle with a null value for contentType, returns an AcademicArticle


torarnet (approved):

ketilaa (approved):
If I understand it correctly, this first PR is about reading these Journal type publication from the database into the new data model, right?

brinxmat:
> If I understand it correctly, this first PR is about reading these Journal type publication from the database into the new data model, right?

Correct

PR #355, opened 2023-02-22
Np 20411 add rights holder

brinxmat:
- Adds a rights holder string to the publication, this is any string purporting to represent data about legal ownership of the data…typically ©<year> <some person(s) or legal entity/entities>


truhacevkir (approved):

ketilaa (approved):
No explicit tests, but I assume it is implicitly tested in existing tests.

PR #357, opened 2023-02-23
make-copy-deep-copy, implemented

truhacevkir:
Should fix shallowCopy problem
P.S: I am using nva stylesheet

### DIFF
--- a/nva-datamodel-java/src/test/java/no/unit/nva/PublicationTest.java
+++ b/nva-datamodel-java/src/test/java/no/unit/nva/PublicationTest.java
@@ -9,6 +9,7 @@ import static no.unit.nva.model.PublicationStatus.PUBLISHED;
 import static no.unit.nva.model.file.FileModelTest.buildAdministrativeAgreement;
 import static no.unit.nva.model.file.FileModelTest.buildNonAdministrativeAgreement;
 import static no.unit.nva.model.file.FileModelTest.randomLegacyFile;
+import static no.unit.nva.model.testing.PublicationGenerator.randomDoi;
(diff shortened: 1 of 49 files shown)


nva-datamodel-java/src/test/java/no/unit/nva/PublicationTest.java
```diff
@@ -117,37 +119,49 @@ void copyReturnsBuilderWithAllDataOfAPublication(Class<?> referenceInstanceType)
...
+    @Test
+    void copyShouldCreateDeepCopyOfPublicationWithoutOverridingOriginalPublicationValuesWhenModifyingCopy() {
+        Publication publication = randomPublication();
+        Publication copy = publication.copy().build();
+        EntityDescription entityDescription = publication.getEntityDescription();
+        entityDescription.setDescription(randomString());
+        copy.setLink(randomUri());
+        copy.setEntityDescription(entityDescription);
+        copy.getEntityDescription().getReference().setDoi(randomDoi());
+        assertThat(publication, not(is(equalTo(copy))));
```
brinxmat, line 132: Possibly use `compareAsObjectNodes` to assert that the actual difference is the expected difference?

brinxmat (changes requested):

nva-datamodel-java/src/test/java/no/unit/nva/PublicationTest.java
brinxmat: Also, you will need to bump the model version

brinxmat (dismissed):

anetteOlli (dismissed):

PR #365, opened 2023-04-03
move role to object

truhacevkir:


nva-datamodel-java/src/main/java/no/unit/nva/model/role/RoleType.java
```diff
@@ -0,0 +1,58 @@
...
+    public static final String DESCRIPTION_FIELD = "description";
+    @JsonProperty(TYPE_FIELD)
+    private final Role type;
+
+    @JsonCreator
+    public RoleType(@JsonProperty(TYPE_FIELD) Role type) {
+        this.type = type;
+    }
+
+    public RoleType createOther(@JsonProperty(DESCRIPTION_FIELD) String description) {
```
ketilaa, line 20: Annotation is probably not needed, and the method could be static?

nva-datamodel-java/src/main/java/no/unit/nva/model/role/RoleType.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @JsonCreator
+    public RoleType fromJson(@JsonProperty(TYPE_FIELD) Role type,
+                             @JsonProperty(DESCRIPTION_FIELD) String description) {
+        return Role.OTHER.equals(type) ? createOther(description) : new RoleType(type);
+    }
+
+    @Deprecated
+    @JsonCreator
+    public RoleType create(Role type) {
```
ketilaa, line 32: Could annotate with @JsonProperty for clarity?

ketilaa (approved):

PR #376, opened 2023-04-21
approvedBy field for Publication

truhacevkir:


nva-datamodel-java/src/main/java/no/unit/nva/model/Agent.java
```diff
@@ -5,6 +5,7 @@
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 @JsonSubTypes({
+    @JsonSubTypes.Type(name = "Approver", value = Approver.class),
```
brinxmat, line 8: We need a quick chat here because we're representing something we already represent elsewhere in a new way.

brinxmat (changes requested):

anetteOlli (approved):

PR #392, opened 2023-05-05
Update pipeline code

brinxmat:
- uses Java 17
- updates java action with gradle cache
- removes cache action since with is called by the java action
- removes unused directives (install jq)


ketilaa (approved):

anetteOlli:
Trengs ikke det å sette java versjon i gradle også?

anetteOlli (approved):

hackon (commented):
Do we need to update more things too?
Like: jitpack?

PR #395, opened 2023-05-10
Publication channels id migration

monaullah:


ketilaa:
Seems like you have some Codacy issues here.

ketilaa (approved):

nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/Publisher.java
```diff
@@ -1,25 +1,34 @@
...
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 public class Publisher implements PublishingHouse {
+
     private final URI id;
 
     @JsonCreator
     public Publisher(@JsonProperty("id") URI id) {
-        this.id = id;
+        this.id = migrateToNewIdIfFound(id, ChannelType.PUBLISHER);
```
brinxmat, line 19: A random thought, if an initial check is placed here (and elsewhere this code is called) so that the migration operation occurs only once, we avoid having to remove the code immediately. E.g.:

```
    this.id = checkIfIsNewStyleUri(id) ? id : migrateToNewIdIfFound(id, ChannelType.PUBLISHER);
```

nva-datamodel-java/src/test/java/no/unit/nva/MigrateSeriesIdTest.java
```diff
@@ -0,0 +1,38 @@
...
+import no.unit.nva.model.contexttypes.utils.ChannelType;
+import org.junit.jupiter.api.Test;
+
+class MigrateSeriesIdTest {
+
+    private static final ChannelType SERIES = ChannelType.SERIES;
+    private static final String OLD_IDENTIFIER_WITHOUT_MAPPING = "111111";
+
+    @Test
+    void shouldReplaceOldSeriesIdWithNewId() {
```
brinxmat, line 21: The three test classes can be reduced to a single migration test that is parameterised and marked as deprecated. The code for this migration should also be marked as deprecated.

brinxmat (changes requested):

nva-datamodel-java/src/test/java/no/unit/nva/MigrateSeriesIdTest.java
monaullah: I did not manage to reduce the tests to a single migration test because i use the constructors for Journal, Publisher and Series in the tests.
Do you have an example of how you would do this?

nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/Publisher.java
monaullah: Clever thought! Added checkIfIsNewStyleIdentifier() checking if the new identifier is a valid UUID.

brinxmat (changes requested):

nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/utils/MigrateChannelIdUtil.java
```diff
@@ -19,39 +22,50 @@ public final class MigrateChannelIdUtil {
...
         var year = UriWrapper.fromUri(id).getPath().getLastPathElement();
         return nonNull(newIdentifier) ? constructNewPublicationChannelId(id, year, newIdentifier, type.pathElement)
                    : id;
     }
 
+    @Deprecated
+    public static boolean checkIfIsNewStyleIdentifier(URI id) {
+        var identifier = UriWrapper.fromUri(id)
+                             .getPath()
+                             .getPathElementByIndexFromEnd(IDENTIFIER_PATH_ELEMENT_INDEX_FROM_END);
```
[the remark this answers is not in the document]
monaullah: We don't want the last path element, the identifier is in the second to last path element:
See example id: https://api.dev.nva.aws.unit.no/publication-channels/journal/_478599_/2017

nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/Journal.java
```diff
@@ -43,9 +46,10 @@ public boolean equals(Object o) {
         return Objects.equals(getId(), journal.getId());
     }
 
-    @JacocoGenerated
-    @Override
-    public int hashCode() {
-        return Objects.hash(getId());
+    private static void validate(URI id) {
+        var stringOfUri = id.toString();
+        if (isNull(stringOfUri) || stringOfUri.isEmpty()) {
```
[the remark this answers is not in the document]
monaullah: _isEmpty()_ returns true if length() is 0, while _isBlank()_ returns true if the string is empty or contains only white space codepoints. Replaced _isEmpty()_  with _isBlank()_  in new commit.

brinxmat (dismissed):

PR #396, opened 2023-05-15
Np 44692 add verified field to contributor

anetteOlli:


truhacevkir (approved):

truhacevkir (approved):

nva-datamodel-java/src/main/java/no/unit/nva/model/ContributorVerificationStatus.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /**
+     * Lookup enum by value.
+     *
+     * @param value value
+     * @return enum
+     */
+    @JsonCreator
+    public static ContributorVerificationStatus parse(String value) {
```
brinxmat, line 36: Call value -> candidate so you can avoid confusion

brinxmat (changes requested):

truhacevkir (dismissed):

nva-datamodel-java/src/main/java/no/unit/nva/model/ContributorVerificationStatus.java
anetteOlli: Fixed

ketilaa (dismissed):

PR #397, opened 2023-05-30
Np 44791 Move Contributor verificationStatus

monaullah:
Moved contributorVerificationStatus from Contributor to Contributor.Identity

### DIFF
--- a/documentation/schema.yaml
+++ b/documentation/schema.yaml
@@ -75,10 +75,10 @@ schema:
     associatedArtifacts:
       type: array
       properties:
-        publishable:
+        empty:
           type: boolean
           example: null
-        empty:
+        publishable:
           type: boolean
           example: null
       example: null
@@ -418,10 +418,10 @@ referencedSchemas:
   AssociatedArtifactList:
     type: array
     properties:
-      publishable:
+      empty:
         type: boolean
         example: null
-      empty:
+      publishable:
         type: boolean
         example: null
     example: null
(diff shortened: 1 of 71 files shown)


anetteOlli (approved):

documentation/schema.yaml
```diff
@@ -780,9 +780,6 @@ referencedSchemas:
         extent:
           type: string
           example: null
-        description:
```
hackon, line 783: It's gone description is?

nva-datamodel-java/src/main/java/no/unit/nva/model/Identity.java
```diff
@@ -68,21 +85,19 @@ public boolean equals(Object o) {
...
     }
 
     public static final class Builder {
+
         private URI id;
         private String name;
         private NameType nameType;
         private String orcId;
 
+        private ContributorVerificationStatus verificationStatus;
```
hackon, line 99: Space above?

hackon (dismissed):

documentation/schema.yaml
monaullah: Yes, Rurik removed in previous commit to main https://github.com/BIBSYSDEV/nva-datamodel-java/commit/7f13a7b513b6cf996f277f7ad59bac2e90ec1308

nva-datamodel-java/src/main/java/no/unit/nva/model/Identity.java
monaullah: Removed in new commit

brinxmat (approved):

PR #411, opened 2023-06-27
Fixed bug with null sequence in testcases

sondrev:


nva-datamodel-java/src/main/java/no/unit/nva/model/EntityDescription.java
```diff
@@ -105,10 +105,13 @@ public void setContributors(List<Contributor> contributors) {
     }
 
     private int compareContributors(Contributor contributor, Contributor otherContributor) {
-        if (contributor.getSequence().equals(otherContributor.getSequence())) {
+        var sequence = contributor.getSequence();
+        var otherSequence = otherContributor.getSequence();
+
+        if (sequence == null ||otherSequence == null || sequence.equals(otherSequence)) {
```
brinxmat, line 111: Objects::isNull

brinxmat (changes requested):

PR #413, opened 2023-08-03
Np 45040 migrate handle to list

truhacevkir:


nva-datamodel-java/src/main/java/no/unit/nva/api/PublicationResponse.java
```diff
@@ -113,13 +116,22 @@ public void setStatus(PublicationStatus status) {
     }
 
     @Override
-    public URI getHandle() {
-        return handle;
+    public Set<URI> getHandles() {
+        return handles;
```
anetteOlli, line 120: pirk: hvis null, returner tom set.

nva-datamodel-java/src/main/java/no/unit/nva/model/Publication.java
```diff
@@ -99,13 +110,22 @@ public void setStatus(PublicationStatus status) {
     }
 
     @Override
-    public URI getHandle() {
-        return handle;
+    public Set<URI> getHandles() {
+        return handles;
```
anetteOlli, line 114: pirk: hvis null...

anetteOlli (approved):

brinxmat:
Why is handle a list? I imagine that we only have one we are interested in, at least in this location in the data model

brinxmat (approved):

PR #429, opened 2023-11-06
Np 44910 publication channel migration prod

monaullah:
Change api host to prod url

### DIFF
--- a/nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/utils/MigrateChannelIdUtil.java
+++ b/nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/utils/MigrateChannelIdUtil.java
@@ -13,7 +13,7 @@ import nva.commons.core.paths.UriWrapper;
 @Deprecated
 public final class MigrateChannelIdUtil {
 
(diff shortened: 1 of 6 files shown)


nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/utils/MigrateChannelIdUtil.java
```diff
@@ -13,7 +13,7 @@
 @Deprecated
 public final class MigrateChannelIdUtil {
 
-    public static final String API_HOST_TEST = "api.test.nva.aws.unit.no";
+    public static final String API_HOST_PROD = "api.nva.aws.unit.no";
```
brinxmat, line 16: `api.nva.unit.no`, I suspect

brinxmat (changes requested):

nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/utils/MigrateChannelIdUtil.java
monaullah: Thanks!

brinxmat (approved):

PR #434, opened 2023-11-23
NP-45719-preserve-scopus-import-author-id

anetteOlli:


nva-datamodel-java/src/main/java/no/unit/nva/model/Identity.java
```diff
@@ -57,6 +61,16 @@ public void setOrcId(String orcId) {
         this.orcId = orcId;
     }
 
+    public List<AdditionalIdentifier> getAdditionalIdentifiers() {
+        return isNull(additionalIdentifiers)
+                   ? List.of()
+                   : additionalIdentifiers;
```
brinxmat, line 67: might have inverted the logic here: `nonNull(additionalIdentifiers) ? additionalIdentifiers : emptyList();`

brinxmat (approved):

PR #437, opened 2023-11-29
adding courseCode to degrees

truhacevkir:


truhacevkir:
We need Confirmed and Unconfirmed course

PR #439, opened 2023-11-30
adding Revision enum to Book

truhacevkir:


brinxmat (approved):
The question is…what does it revise?

truhacevkir:
> The question is…what does it revise?

It does not revise, but indicate that resource has been changed/ modified, etc. has some changes from original resource.

PR #441, opened 2023-12-05
publishedDate, createdDate, modifiedDate, publicationDate comments

anetteOlli:


truhacevkir (approved):

elkebab (dismissed):
LGTM 👍

nva-datamodel-java/src/main/resources/publication-ontology.ttl
```diff
@@ -844,7 +844,7 @@ nva:JournalReview a rdfs:Class ;
   rdfs:subClassOf nva:Journal .
 
 nva:PublicationDate a rdfs:Class ;
-  rdfs:comment "" ;
+  rdfs:comment "When the publication was first published in any system or form" ;
```
brinxmat, line 847: I think it is "The date the thing was first published for public consumption", using the word "publication" is problematic since it relates closely to "Publication", which PublicationDates are never directly associated with.

nva-datamodel-java/src/main/resources/publication-ontology.ttl
```diff
@@ -1343,7 +1343,7 @@ nva:concertProgramme a rdf:Property ;
   rdfs:range nva:MusicalWorkPerformance .
 
 nva:publishedDate a rdf:Property ;
-  rdfs:comment "This represents meta-metadata about when the Publication description changed status to Published" ;
+  rdfs:comment "This represents meta-metadata about when the Publication description changed status to Published in this system" ;
```
brinxmat, line 1346: "The date that the thing was originally published for public consumption in the documenting system" avoids details

nva-datamodel-java/src/main/resources/publication-ontology.ttl
```diff
@@ -1644,7 +1644,7 @@ nva:name a rdf:Property ;
   rdfs:range xsd:string .
 
 nva:modifiedDate a rdf:Property ;
-  rdfs:comment "" ;
+  rdfs:comment "When the publication was last modified in this system" ;
```
brinxmat, line 1647: in the documenting system

nva-datamodel-java/src/main/resources/publication-ontology.ttl
brinxmat, line 847: This is a thankless task since we need to agree…and we will all have very different ideas about what words mean…it should have been done at the beginning and maintained. Thanks for picking this up!

brinxmat (changes requested):

nva-datamodel-java/src/main/resources/publication-ontology.ttl
anetteOlli: Your suggestion is so much better than mine, thank you.

brinxmat (approved):

PR #450, opened 2024-01-04
making it possible to create unpublishable file that is not Adm.agreem.

truhacevkir:


brinxmat (approved):
- Fix the codacy issues
- Do we need tests for this? I suspect we do

PR #453, opened 2024-01-09
Removing labels from organization

truhacevkir:


nva-datamodel-java/src/test/java/OrganizationTest.java
```diff
@@ -0,0 +1,21 @@
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import no.unit.nva.commons.json.JsonUtils;
+import no.unit.nva.model.Organization;
+import org.junit.jupiter.api.Test;
+
+public class OrganizationTest {
```
brinxmat, line 6: Mark as `@Deprecated` since we don't want this test over time

brinxmat (changes requested):

PR #460, opened 2024-02-06
Np 45941 add unpublish comment

torbjokv:


anetteOlli (approved):

nva-datamodel-java/src/main/java/no/unit/nva/model/PublicationUnpublishNote.java
```diff
@@ -0,0 +1,8 @@
+package no.unit.nva.model;
+
+import com.fasterxml.jackson.annotation.JsonTypeName;
+
+@JsonTypeName("PublicationUnpublishNote")
+public record PublicationUnpublishNote(String note) implements PublicationNoteBase {
```
brinxmat, line 6: `UnpublishingNote`?

brinxmat (changes requested):
I'd change the name of the object to keep names short — otherwise merge away!

brinxmat (approved):

PR #467, opened 2024-02-18
randomNonDegree publication and renaming republish operation

sondrev:


nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/PublicationGenerator.java
```diff
@@ -71,6 +77,18 @@ public static Publication randomPublication(Class<?> publicationInstanceClass) {
 
         return buildRandomPublicationFromInstance(publicationInstanceClass);
     }
+    public static Publication randomPublicationNonDegree() {
+        var degrees = Set.of(DegreeMaster.class,
+                             DegreeBachelor.class,
+                             DegreePhd.class,
+                             DegreeLicentiate.class,
+                             OtherStudentWork.class);
```
sondrev, line 85: In terms of accessRights ThesisCurators are the only ones supposed to edit an Degree - does this include "OtherStudentWork" - it inherits from BaseDegree

nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/PublicationGenerator.java
```diff
@@ -71,6 +77,18 @@ public static Publication randomPublication(Class<?> publicationInstanceClass) {
...
+        var nonDegrees =
+            PublicationInstanceBuilder.listPublicationInstanceTypes()
+                .stream().filter(i -> !degrees.contains(i)
+            ).collect(Collectors.toList());
```
sondrev, line 89: Would love to use a cleaner approch, but struggled with generics because listPublicationInstanceTypes() returns Class<?>.

brinxmat:
```
    public static Publication randomPublicationNonDegree() {
        var nonDegrees = PublicationInstanceBuilder.listPublicationInstanceTypes()
                .stream()
                .filter(not(PublicationGenerator::isDegree))
                .toList();
        return randomPublication(randomElement(nonDegrees));
    }

    private static boolean isDegree(Class<?> subClass) {
        return subClass.isAssignableFrom(DegreeBase.class);
    }
```

brinxmat (changes requested):

nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/PublicationGenerator.java
brinxmat: Because we want the complement of the set of Degrees, we would have to do some jiggery pokery to get all classpath classes, so this is likely the best approach

PR #474, opened 2024-03-06
feat(publication): Add CuratingInstitutions to publication #NP-46344

torbjokv:


nva-datamodel-java/src/main/resources/publicationContext.json
```diff
@@ -179,5 +179,10 @@
   "venues": {
     "@id": "venue",
     "@container": "@set"
+  },
+  "curatingInstitutions": {
+    "@id": "curatingInstitution",
```
torbjokv, line 184: @brinxmat "@id" should maybe be organization?

nva-datamodel-java/src/main/java/no/unit/nva/model/Publication.java
```diff
@@ -390,6 +394,14 @@ public boolean satisfiesFindableDoiRequirements() {
         return FindableDoiRequirementsValidator.meetsFindableDoiRequirements(this);
     }
 
+    public Set<URI> getCuratingInstitutions() {
+        return nonNull(curatingInstitutions) ? curatingInstitutions : Collections.emptySet();
```
sondrev, line 398: Why not just return null? There is a difference "not set" and "no curating institutions"

sondrev (changes requested):
1. see coment
2. bump version number for datamodel
3. You need to update PublicationResponse aswell

nva-datamodel-java/src/main/java/no/unit/nva/model/Publication.java
brinxmat: @sondrev For the sake of the receiver, in serialization

brinxmat (dismissed):

PR #478, opened 2024-03-12
Support contributorOrganizations in context

brinxmat:
- Fixes context for later processing


nva-datamodel-java/src/test/java/no/unit/nva/PublicationContextTest.java
```diff
@@ -82,10 +82,12 @@ void shouldWriteEveryNvaClassAsFragmentUri() {
     }
 
     @Test
-    void shouldDefineEveryArrayPropertyAsContainerSet() {
```
brinxmat, line 85: This test was weird, I changed  it so it looks only at the definitions in the model without regard to having 100% equality with the context object

brinxmat (commented):
OBS: If you filter the JSON files, you won't see the context.json changes!

anetteOlli (approved):

truhacevkir (approved):

PR #483, opened 2024-05-06
adding getters to UnconfirmedPlace

truhacevkir:
I need to compare these fields to UnconfirmedPlace when running import in order to be able to update one of them when missing.


nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/place/UnconfirmedPlace.java
```diff
@@ -39,4 +39,12 @@ public boolean equals(Object o) {
...
         return Objects.hash(label, country);
     }
+
+    public String getLabel() {
+        return label;
+    }
+
+    public String getCountry() {
+        return country;
+    }
```
truhacevkir, line 49: This is the only change - adding getters to `UnconfirmedPlace`

anetteOlli (approved):

PR #485, opened 2024-05-21
NP-46774-expand-publisher-version-to-support-more-types

anetteOlli:
According to the testers, we need to support "draft", "updatedVersion", and "submittedVersion"

Also removed backwards (and already flagged as deprecated) functionality to parse booleans as accepted / published version.

### DIFF
--- a/buildSrc/src/main/groovy/nvadatamodel.java-conventions.gradle
+++ b/buildSrc/src/main/groovy/nvadatamodel.java-conventions.gradle
@@ -10,7 +10,7 @@ plugins {
 group 'com.github.bibsysdev'
 
 
-version '0.21.20'
+version '0.22.01'
 
 repositories {
     mavenCentral()
--- a/nva-datamodel-java/src/main/java/no/unit/nva/model/associatedartifacts/file/PublisherVersion.java
+++ b/nva-datamodel-java/src/main/java/no/unit/nva/model/associatedartifacts/file/PublisherVersion.java
@@ -9,7 +9,7 @@ import nva.commons.core.StringUtils;
 
 public enum PublisherVersion {
     PUBLISHED_VERSION("PublishedVersion"),
-    DRAFT("draft"),
+    DRAFT("Draft"),
     SUBMITTED_VERSION("SubmittedVersion"),
     UPDATED_VERSION("UpdatedVersion"),
     ACCEPTED_VERSION("AcceptedVersion");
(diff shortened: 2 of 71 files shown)


buildSrc/src/main/groovy/nvadatamodel.java-conventions.gradle
```diff
@@ -10,7 +10,7 @@ plugins {
 group 'com.github.bibsysdev'
 
 
-version '0.21.20'
+version '0.22.01'
```
anetteOlli, line 13: No longer 100 % backwards compatible.

nva-datamodel-java/src/main/java/no/unit/nva/model/associatedartifacts/file/PublisherVersion.java
```diff
@@ -2,18 +2,19 @@
...
 import static java.util.stream.Collectors.joining;
 import com.fasterxml.jackson.annotation.JsonValue;
 import no.unit.nva.model.Revision;
+import nva.commons.core.StringUtils;
 
 public enum PublisherVersion {
     PUBLISHED_VERSION("PublishedVersion"),
+    DRAFT("draft"),
+    SUBMITTED_VERSION("SubmittedVersion"),
+    UPDATED_VERSION("UpdatedVersion"),
```
anetteOlli, line 14: Names copied straight from the task. Feel free to suggest other names.

nva-datamodel-java/src/test/java/no/unit/nva/model/file/FileModelTest.java
```diff
@@ -293,6 +258,22 @@ private static File.Builder admAgreementBuilder() {
                    .withName(randomString());
     }
 
+    private static String generateNewFile() {
```
anetteOlli, line 261: Checkstyle.

nva-datamodel-java/src/test/java/no/unit/nva/model/file/FileModelTest.java
```diff
@@ -293,6 +233,22 @@ private static File.Builder admAgreementBuilder() {
...
```
anetteOlli, line 236: Checkstyle plus changed publisherAuthority from boolean to publisherVersion enum.

nva-datamodel-java/src/main/java/no/unit/nva/model/associatedartifacts/file/PublisherVersion.java
truhacevkir: I would go for "Draft" instead of "draft", so it matches the rest of enums

truhacevkir (approved):

brinxmat (approved):

nva-datamodel-java/src/main/java/no/unit/nva/model/associatedartifacts/file/PublisherVersion.java
brinxmat: Good catch :)

truhacevkir:
Pratet med Terje, det er ikke nødvendig å utvide datamodellen.

PR #488, opened 2024-05-29
NP-46821 Add copy method with builder to files

joachimjorgensen:
Add copy-method to files. Had to move `publishedDate` out of `PublishedFile` and into `File` to reuse `Builder`-class. `publishedDate` is only set in `PublishedFile` constructor.

Sneakily added `getUploadedDate()` to `UploadDetails`.


anetteOlli (dismissed):
Remember to bump nva-datamodel in nva-fetch-doi, publication-api and download-file

PR #489, opened 2024-05-31
implementing and adding Duration to MovingPicture

truhacevkir:
Implementing Duration and adding it to MovingPicture in order to migrate duration for Cristin movies and other artistic types.

Implementing three classes:

NullDuration = when duration is missing.
UndefinedDuration = when duration is present but we can not identify time unit.
DefinedDuration = when duration is present and we can identify time unit.

### DIFF
--- a/nva-datamodel-java/src/main/java/no/unit/nva/model/time/duration/DefinedDuration.java
+++ b/nva-datamodel-java/src/main/java/no/unit/nva/model/time/duration/DefinedDuration.java
@@ -13,14 +13,14 @@ public final class DefinedDuration implements Duration {
     public static final String HOURS_FIELD = "hours";
     public static final String DAYS_FIELD = "days";
     public static final String WEEKS_FIELD = "weeks";
-    private final String minutes;
-    private final String hours;
-    private final String days;
-    private final String weeks;
+    private final int minutes;
+    private final int hours;
+    private final int days;
+    private final int weeks;
 
     @JsonCreator
-    private DefinedDuration(@JsonProperty(MINUTES_FIELD) String minutes, @JsonProperty(HOURS_FIELD) String hours,
-                            @JsonProperty(DAYS_FIELD) String days, @JsonProperty(WEEKS_FIELD) String weeks) {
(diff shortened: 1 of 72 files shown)


nva-datamodel-java/src/main/java/no/unit/nva/model/time/duration/DefinedDuration.java
```diff
@@ -0,0 +1,105 @@
+package no.unit.nva.model.time.duration;
```
ketilaa, line 1: There is an ISO standard for duration too.

So,
```
"duration": {
  "type": "DefinedDuration",
  "minutes": "35",
  "hours": "2",
  "days": "0",
  "weeks": "0"
}
```
could be written as:
```
"duration": "PT2H35M"
```
Anyway, if it is a well-defined duration, `minutes`, `hours`, `days` and `weeks` should be integers.

nva-datamodel-java/src/test/java/no/unit/nva/LicenseMigrationTest.java
```diff
@@ -4,6 +4,7 @@
 import java.util.stream.Stream;
 import no.unit.nva.commons.json.JsonUtils;
 import no.unit.nva.model.associatedartifacts.file.File;
+import no.unit.nva.model.time.duration.DefinedDuration;
```
ketilaa, line 7: Unused?

brinxmat (dismissed):

nva-datamodel-java/src/main/java/no/unit/nva/model/time/duration/DefinedDuration.java
```diff
@@ -0,0 +1,105 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import java.util.Objects;
+import nva.commons.core.JacocoGenerated;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public final class DefinedDuration implements Duration {
```
anetteOlli, line 10: Should we document this in onthologies?

anetteOlli (dismissed):

PR #490, opened 2024-06-06
np-46885-degree-thesis-fixes

ketilaa:
The datamodel repository contains PublicationGenerator class to help generate random publications during testing. For testing special access requirements for degree thesis, and to avoid these publication instance types when testing different scenarios, a method was added to generate random non-degree publications. However, this was based on instance being a subclass of DegreeBase, and that is not sufficient for the special access requirements that only applies to DegreePhd, DegreeMaster, DegreeBachelor and DegreeLicentiate.

In this PR the previous method is deprecated and marked for deletion, while I added two new methods to generate either random degree thesis or random non-degree thesis publications.

### DIFF
--- a/nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/PublicationGenerator.java
+++ b/nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/PublicationGenerator.java
@@ -9,6 +9,7 @@ import static no.unit.nva.testutils.RandomDataGenerator.randomInstant;
 import static no.unit.nva.testutils.RandomDataGenerator.randomInteger;
 import static no.unit.nva.testutils.RandomDataGenerator.randomString;
 import java.net.URI;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
 import java.util.UUID;
@@ -32,11 +33,6 @@ import no.unit.nva.model.Username;
 import no.unit.nva.model.funding.Funding;
 import no.unit.nva.model.funding.FundingBuilder;
 import no.unit.nva.model.funding.MonetaryAmount;
-import no.unit.nva.model.instancetypes.degree.DegreeBachelor;
-import no.unit.nva.model.instancetypes.degree.DegreeBase;
-import no.unit.nva.model.instancetypes.degree.DegreeLicentiate;
-import no.unit.nva.model.instancetypes.degree.DegreeMaster;
-import no.unit.nva.model.instancetypes.degree.DegreePhd;
 import no.unit.nva.model.testing.associatedartifacts.AssociatedArtifactsGenerator;
 import nva.commons.core.JacocoGenerated;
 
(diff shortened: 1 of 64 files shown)


nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/PublicationGenerator.java
```diff
@@ -82,10 +91,33 @@ public static Publication randomPublicationNonDegree() {
         return randomPublication(randomElement(nonDegrees));
     }
 
+    public static Publication randomPublicationNonDegreeThesis() {
```
torbjokv, line 94: do we need this?

torbjokv: maybe add javadoc comment that explains difference

nva-datamodel-testutils/src/test/java/no/unit/nva/model/testing/PublicationGeneratorTest.java
```diff
@@ -31,6 +35,11 @@ public static Stream<Class<?>> publicationInstanceProvider() {
         return PublicationInstanceBuilder.listPublicationInstanceTypes().stream();
     }
 
+    private static final Set<Class<?>> DEGREE_THESIS_INSTANCE_TYPES = Set.of(DegreePhd.class,
```
brinxmat, line 38: This should be moved up above the non-statiic fields

brinxmat (changes requested):

torbjokv (approved):

nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/PublicationGenerator.java
```diff
@@ -74,16 +74,18 @@ public static Publication randomPublication(Class<?> publicationInstanceClass) {
...
 
-    public static Publication randomPublicationNonDegree() {
-        var nonDegrees = PublicationInstanceBuilder.listPublicationInstanceTypes()
-                             .stream()
-                             .filter(not(PublicationGenerator::isDegree))
-                             .toList();
-        return randomPublication(randomElement(nonDegrees));
+    public static Publication ofInstanceClasses(Class<?>... publicationInstanceClasses) {
+        var publicationInstanceClass = randomElement(publicationInstanceClasses);
+        return randomPublication(publicationInstanceClass);
```
torbjokv, line 79: maybe rewrite randomPublication instead of adding ofInstanceClasse

brinxmat: I think it is correct as is, but perhaps `from` not `of` since we then get`PublicationGenerator.fromInstanceClasses(Class<?>...targetClasses)`. (I'd change the param name too).

nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/PublicationGenerator.java
```diff
@@ -78,44 +74,18 @@ public static Publication randomPublication(Class<?> publicationInstanceClass) {
...
-    private static boolean isDegree(Class<?> subClass) {
-        return DegreeBase.class.isAssignableFrom(subClass);
-    }
-
-    private static boolean isDegreeThesis(Class<?> subClass) {
-        return subClass.equals(DegreePhd.class) ||
-               subClass.equals(DegreeMaster.class) ||
-               subClass.equals(DegreeBachelor.class) ||
-               subClass.equals(DegreeLicentiate.class);
+    public static Publication notOfInstanceClasses(Class<?>... publicationInstanceClasses) {
```
brinxmat, line 82: `fromInstanceClassesExcluding(Class<?>... excludeClasses)`

brinxmat (changes requested):

nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/PublicationGenerator.java
ketilaa: refactored

nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/PublicationGenerator.java
ketilaa: Followed @brinxmat 's approach.

nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/PublicationGenerator.java
ketilaa: refactored

PR #491, opened 2024-06-10
adding partOf getter to media contribution

truhacevkir:
I need this getter to be able to update partOf value when merging two publications

### DIFF
--- a/nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/MediaContribution.java
+++ b/nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/MediaContribution.java
@@ -70,13 +70,13 @@ public class MediaContribution implements PublicationContext {
         return Objects.equals(getMedium(), that.getMedium())
                 && getFormat() == that.getFormat()
                 && Objects.equals(getDisseminationChannel(), that.getDisseminationChannel())
-                && Objects.equals(partOf, that.partOf);
+                && Objects.equals(getPartOf(), that.getPartOf());
     }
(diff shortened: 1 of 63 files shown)


nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/MediaContribution.java
```diff
@@ -53,6 +53,10 @@ public String getDisseminationChannel() {
         return disseminationChannel;
     }
 
+    public SeriesEpisode getPartOf() {
+        return partOf;
+    }
+
     @JacocoGenerated
```
truhacevkir, line 60: This is the only change

torbjokv (approved):

nva-datamodel-java/src/main/java/no/unit/nva/model/contexttypes/MediaContribution.java
```diff
@@ -53,6 +53,10 @@ public String getDisseminationChannel() {
         return disseminationChannel;
     }
 
+    public SeriesEpisode getPartOf() {
```
anetteOlli, line 56: Do you need to update equals and hash method?

torbjokv: I see partOf is there, but should probably use getPartOf() to make it **beautiful**

anetteOlli (approved):

torbjokv (approved):

PR #492, opened 2024-06-10
Np 46791 add import details to publication

joachimjorgensen:
Add field `List<ImportDetail> importDetails;` to the Publication model as a mean to identify imported publications. A publication can be imported in multiple instances (e.g. first Brage migration, and later Cristin import), therefore it is a list of import details.

Add enum `ImportSource` that can be expanded when new rounds of importing/migrating are conducted.

### DIFF
--- a/nva-datamodel-java/src/main/java/no/unit/nva/model/ImportDetail.java
+++ b/nva-datamodel-java/src/main/java/no/unit/nva/model/ImportDetail.java
@@ -1,0 +1,5 @@
+package no.unit.nva.model;
+
+import java.time.Instant;
+
+public record ImportDetails(Instant date, ImportSource source) { }
--- a/nva-datamodel-java/src/main/java/no/unit/nva/model/ImportDetail.java
+++ b/nva-datamodel-java/src/main/java/no/unit/nva/model/ImportDetail.java
@@ -1,5 +1,7 @@
 package no.unit.nva.model;
 
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import java.time.Instant;
 
-public record ImportDetail(Instant date, ImportSource source) { }
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public record ImportDetail(Instant importDate, ImportSource source) { }
--- a/nva-datamodel-java/src/main/java/no/unit/nva/model/ImportDetail.java
+++ b/nva-datamodel-java/src/main/java/no/unit/nva/model/ImportDetail.java
@@ -2,4 +2,4 @@ package no.unit.nva.model;
 
 import java.time.Instant;
 
-public record ImportDetails(Instant date, ImportSource source) { }
+public record ImportDetail(Instant date, ImportSource source) { }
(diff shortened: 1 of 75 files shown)


nva-datamodel-java/src/main/java/no/unit/nva/model/ImportDetail.java
```diff
@@ -0,0 +1,5 @@
+package no.unit.nva.model;
+
+import java.time.Instant;
+
+public record ImportDetail(Instant date, ImportSource source) { }
```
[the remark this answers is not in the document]
truhacevkir: I would also add following annotation to this record:
`@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")`

So it serializes like this:
```
{
    "type" : "ImportDetail",
    "date" : "2024-06-11T07:01:51.427190Z",
    "source" : "BRAGE"
  }
```
Instead of:
```
{
    "date" : "2024-06-11T07:01:51.427190Z",
    "source" : "BRAGE"
  }
```

nva-datamodel-java/src/main/java/no/unit/nva/model/Publication.java
```diff
@@ -402,6 +405,14 @@ public void setCuratingInstitutions(Set<URI> curatingInstitutions) {
...
 
+    public List<ImportDetail> getImportDetails() {
+        return nonNull(importDetails) ? importDetails : Collections.emptyList();
+    }
+
+    public void setImportDetails(List<ImportDetail> importDetails) {
+        this.importDetails = importDetails;
+    }
+
     private void verifyStatusTransition(PublicationStatus nextStatus)
```
truhacevkir, line 416: Do we need `addImportDetail(ImportDetail importDetail)` method?

anetteOlli:
I have one thing to say about this: AUDIT TRAIL!!!


https://en.wikipedia.org/wiki/Audit_trail

nva-datamodel-java/src/main/java/no/unit/nva/model/ImportDetail.java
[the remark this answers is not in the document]
joachimjorgensen: Good points! Will do

nva-datamodel-java/src/main/java/no/unit/nva/model/Publication.java
```diff
@@ -402,6 +409,30 @@ public void setCuratingInstitutions(Set<URI> curatingInstitutions) {
...
+        if (importDetail == null) {
+            return;
+        }
+
+        if (isNull(importDetails)) {
+            importDetails = new ArrayList<>();
+        }
+
+        importDetails.add(importDetail);
+    }
```
truhacevkir, line 434: You are using both isNull() and something == null, maybe use isNull() only :)

truhacevkir (dismissed):

nva-datamodel-java/src/main/java/no/unit/nva/model/Publication.java
joachimjorgensen: Wops, will fix

PR #494, opened 2024-06-13
NP-46791 Update ontology and publicationResponse for ImportDetail

joachimjorgensen:


documentation/schema.yaml
```diff
@@ -1281,14 +1281,14 @@ referencedSchemas:
       source:
         type: string
         format: uri
+      fundingAmount:
```
brinxmat, line 1284: Did this not get updated with ImportDetails on build?

brinxmat (approved):

documentation/schema.yaml
brinxmat: It was there from before :D

PR #497, opened 2024-06-19
adding archive to importSource

truhacevkir:
Refactoring ImportSource to Object containing source and archive


nva-datamodel-java/src/main/java/no/unit/nva/model/ImportDetail.java
```diff
@@ -2,6 +2,16 @@
...
-public record ImportDetail(Instant importDate, ImportSource source) { }
+public record ImportDetail(Instant importDate, ImportSource importSource) {
+
+    public static ImportDetail fromSource(Source source, Instant importDate) {
+        return new ImportDetail(importDate, ImportSource.fromSource(source));
+    }
+
+    public static ImportDetail fromBrageArchive(String archive, Instant importDate) {
+        return new ImportDetail(importDate, ImportSource.fromBrageArchive(archive));
+    }
```
truhacevkir, line 16: Convenient with these two methods?

So we just run 
`ImportDetail.fromBrageArchive(archive, date)` 
instead of 
`new ImportDetail(date, new ImportSource(source, archive)`

buildSrc/src/main/groovy/nvadatamodel.java-conventions.gradle
```diff
@@ -10,7 +10,7 @@ plugins {
 group 'com.github.bibsysdev'
 
 
-version '0.22.0'
+version '0.22.1'
```
joachimjorgensen, line 13: It probably doesn't matter, but is there a reason we have gone from 0.21.29 to 0.22.0 instead of 0.21.30?

truhacevkir: Anette meant this is a new "major" version since it is no longer compatible with older versions implementations, since we removed `publisherAuthority`

nva-datamodel-java/src/main/resources/publication-ontology.ttl
joachimjorgensen, line 1: At line 1516 you have to remove nva:ImportDetail, and instead create a new Class ImportSource, which references ImportDetail. And at the bottom there is a property importDate taht might have to be copied for importSource

nva-datamodel-java/src/main/resources/publication-ontology.ttl
```diff
@@ -18,16 +18,16 @@ nva:Exhibition a rdfs:Class ;
   rdfs:subClassOf nva:ArtisticDesignSubtype .
 
 nva:Funding a rdfs:Class ;
-  rdfs:comment "A source of financial support" ;
+  rdfs:comment "A importSource of financial support" ;
```
joachimjorgensen, line 21: Undo all renaming relating to funding

buildSrc/src/main/groovy/nvadatamodel.java-conventions.gradle
joachimjorgensen: Ah, then I agree 👍

nva-datamodel-java/src/main/resources/publication-ontology.ttl
joachimjorgensen, line 1: You also have to add this at the bottom, if my crash course by Rurik serves me right:
```nva:importSource a rdf:Property ;
 rdfs:comment "The source of an imported Publication" ;
 rdfs:label "Import source" ;
 rdfs:isDefinedBy nva: ;
 rdfs:domain nva:ImportDetail ;
 rdfs:range xsd:ImportSource .```

Maybe we also have to define properties on ImportSource? Not sure

truhacevkir: nva:Source a rdf:Property ;
 rdfs:comment "The source of an imported Publication" ;
 rdfs:label "Import source" ;
 rdfs:isDefinedBy nva: ;
 rdfs:domain nva:ImportDetail ;
 rdfs:range xsd:ImportSource .

joachimjorgensen (approved):

PR #499, opened 2024-07-01
feature(additionalIdentifiers): Add support for CristinIdentifier, HandleIdentifier and ScopusIdentifier #NP-47114

torbjokv:


nva-datamodel-java/src/main/java/no/unit/nva/model/CristinIdentifier.java
```diff
@@ -0,0 +1,7 @@
+package no.unit.nva.model;
+
+public record CristinIdentifier(String sourceName, String value) implements AdditionalIdentifierBase {
+
+    static final String TYPE = "CristinIdentifier";
+
```
brinxmat, line 6: What is the sourceName here? The result uri? If so, maybe URI

nva-datamodel-java/src/test/java/no/unit/nva/PublicationTest.java
```diff
@@ -121,6 +121,31 @@ void publicationReturnsValidPublicationWhenInputIsValid(Class<?> instanceType) t
         writePublicationToFile(instanceType, expected);
     }
 
+    @Test
+    void additionalIdentifierShouldSupportSourceAlias() throws Exception {
+        String payload = """
```
brinxmat, line 126: You don't like vars?

brinxmat (changes requested):
-  What happens with already existing values? do we care, or just fix in migration code?
- I guess we have robust round-tripping to/from JSON, so otherwise OK.

nva-datamodel-java/src/main/java/no/unit/nva/model/AdditionalIdentifier.java
```diff
@@ -1,78 +1,8 @@
...
+public record AdditionalIdentifier(String sourceName, String value) implements AdditionalIdentifierBase {
 
-    // TODO: Remove alias when data has been migrated
-    @Deprecated
-    @JsonAlias("source")
-    @JsonProperty("sourceName")
-    private String sourceName;
-    @JsonProperty("value")
-    private String value;
+    static final String TYPE = "AdditionalIdentifier";
```
truhacevkir, line 6: Jeg kunne også se for meg hjelpemdetoder som setter strenger med faster verdier for oss:

```
        AdditionalIdentifier.fromScopus(randomString());
        AdditionalIdentifier.fromCristin(randomString());
        AdditionalIdentifier.fromBrage(randomString(), randomString())
```

nva-datamodel-java/src/main/java/no/unit/nva/model/HandleIdentifier.java
```diff
@@ -0,0 +1,27 @@
...
+
+    @Override
+    public String sourceName() {
+        return source.toString();
+    }
+
+    @Override
+    public String value() {
+        return uri.toString();
+    }
```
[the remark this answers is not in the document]
torbjokv: I think this is ok? should not add Handle if that is the case

nva-datamodel-java/src/main/java/no/unit/nva/model/AdditionalIdentifier.java
torbjokv: hmm I agree that its convenient and could help making more consistant data, but not sure where to add it. suggest adding it when we need it.

nva-datamodel-java/src/main/java/no/unit/nva/model/CristinIdentifier.java
torbjokv: suggest doing SourceName class for all the new identifiers. source is either cristin, brage and  probably ova

nva-datamodel-java/src/main/java/no/unit/nva/model/HandleIdentifier.java
```diff
@@ -0,0 +1,27 @@
...
+
+public record HandleIdentifier(@JsonIgnore SourceName source, @JsonIgnore URI uri)
+    implements AdditionalIdentifierBase {
+
+    static final String TYPE = "HandleIdentifier";
+
+    @JsonCreator
+    public HandleIdentifier(@JsonProperty("sourceName") String sourceName, @JsonProperty("value") String value) {
+        this(new SourceName(sourceName), URI.create(value));
+    }
```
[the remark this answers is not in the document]
torbjokv: made this constructor private, should only be used for json deserialization

nva-datamodel-java/src/main/java/no/unit/nva/model/SourceName.java
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * A record representing a source name.
+ *
+ * @param system the system name of the source. e.g. "brage"
+ * @param instanceName the instance name of the source. e.g. "194.0.0.0"
+ */
+public record SourceName(String system, String instanceName) {
+
+    @JsonValue
```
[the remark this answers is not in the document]
torbjokv: I think SourceName could be relevant for multiple systems like nva, brage, cristin(import), sentralimport/nva, etc

torbjokv:
added a validate method in SourceName to ensure trimmed and lowercase source names

torbjokv:
> * What happens with already existing values? do we care, or just fix in migration code?

plan to deal with it later in migration code

torbjokv:
> * I guess we have robust round-tripping to/from JSON, so otherwise OK.

yes, most of my intermediate errors was caught by the round trip with exception of the source alias that I added a test for. One thing that is not tested for is the output json format, but it's relatively easy to spot changes in the generated files.

torbjokv:
Changed Identifier to use AdditionalIdentifier instead of AdditionalIdentifierBase

nva-datamodel-java/src/main/java/no/unit/nva/model/AdditionalIdentifier.java
torbjokv: added fromBrage and fromCristin helpers to SourceName.

`new HandleIdentifier(SourceName.fromBrage("20754.0.0.0"), randomUri())`

and made singe parameter SourceName constructor package private

nva-datamodel-java/src/test/java/no/unit/nva/PublicationTest.java
```diff
@@ -115,15 +115,15 @@ void publicationReturnsValidPublicationWhenInputIsValid(Class<?> instanceType) t
         Diff diff = JAVERS.compare(expected, roundTripped);
         assertThat(publication, doesNotHaveEmptyValuesIgnoringFields(IGNORE_LIST));
         assertThat(publication, doesNotHaveEmptyValuesIgnoringFields(IGNORE_LIST));
-        assertEquals(expected, roundTripped);
+        //assertEquals(expected, roundTripped);
```
brinxmat, line 118: En stygg kommentar

nva-datamodel-testutils/src/main/java/no/unit/nva/model/testing/PublicationGenerator.java
```diff
@@ -211,7 +218,7 @@ public static Organization randomOrganization() {
 
     private static Publication buildRandomPublicationFromInstance(Class<?> publicationInstanceClass) {
         return new Builder()
-                   .withIdentifier(SortableIdentifier.next())
+                   .withIdentifier(SortableIdentifier.  next())
```
brinxmat, line 221: Space, the final frontier

brinxmat (changes requested):

brinxmat (approved):

